PR #50, opened 2023-10-20
feat: add mysql image

dlen:
Hello!

I think mysql image is missing and it could come handy (code is very much the same as psql image). Let me know what you think!

Thanks for maintaining all these images!


src/mysql/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use testcontainers::{core::WaitFor, Image};
+
+const NAME: &str = "mysql";
+const TAG: &str = "8.1";
+
+#[derive(Debug)]
+pub struct Mysql {
+    env_vars: HashMap<String, String>,
+}
```
DDtKey, line 11: Could you add some basic documentation? We've not added documentation for all our images yet, but it's planned & would be nice to have for new images.

For example, take a look at this one:
https://github.com/testcontainers/testcontainers-rs-modules-community/blob/300f744ee649465e085668095fbefbabd146aebc/src/rabbitmq/mod.rs#L6-L31

I found useful to add at least references.

DDtKey (commented):
Thanks for your contribution!

Yes, it makes sense to add `mysql`.

src/mysql/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn tag(&self) -> String {
+        TAG.to_owned()
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![
+            WaitFor::message_on_stderr("/usr/sbin/mysqld: ready for connections."),
+            WaitFor::Duration {
+                length: Duration::new(10, 0),
+            },
```
DDtKey, line 39: Do we really need it? Probably one more message can indicate it's ready?
If so, consider using https://docs.rs/testcontainers/0.15.0/testcontainers/core/enum.WaitFor.html#method.seconds

src/mysql/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+        )
+        .unwrap();
+
+        let first_row = conn.query_first("SELECT 1+1;").unwrap();
+        assert_eq!(first_row, Some(2));
+
+        let first_column: i32 = first_row.unwrap();
+        assert_eq!(first_column, 2);
+    }
+
```
DDtKey, line 96: Let's remove this test, mapped-port is basic functionality of core-crate `testcontainers`, while `free_local_port` impl is actually a bit flaky. 

It's just an inheritance for postgres image and will be removed later

```suggestion
```

src/mysql/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+        assert_eq!(first_row, Some(String::from("8.0.34")));
+    }
+
+    #[must_use]
+    fn free_local_port() -> u16 {
+        std::net::TcpListener::bind((std::net::Ipv4Addr::LOCALHOST, 0))
+            .and_then(|listener| listener.local_addr())
+            .map(|addr| addr.port())
+            .expect("free port not found")
+    }
```
DDtKey, line 119:
```suggestion
```

src/mysql/mod.rs
dlen: Yeah, I don't like it either. Apparently mysql container restarts once due to (my best guess) password setup or some other janitorial work they do in the entrypoint. We could detect if at least the ready for connections message is output twice instead of the WaitFor::Duration I was in a rush so this is the fastest thing I managed to come up with this morning. I can investigate a bit further

src/mysql/mod.rs
dlen: added docs

src/mysql/mod.rs
dlen: I thought about adding a healthcheck something I have done in other [projects](https://github.com/passbolt/passbolt_docker/blob/58b0ea821eec8bb55e0e3e1b25ca8b8e819b737e/spec/docker_runtime/runtime_spec.rb#L21) but I think I would need to open a PR on the testcontainers repo as I don't see any method to define it, right?

src/mysql/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// This module is based on the officlal [`MySQL docker image`].
+///
+/// # Example
+/// ```
+/// use testcontainers::clients;
+/// use testcontainers_modules::mysql;
+///
+/// let docker = clients::Cli::default();
+/// let mysql_instance = docker.run(mysql::Mysql);
```
DDtKey, line 20:
```suggestion
/// let mysql_instance = docker.run(mysql::Mysql::default());
```

src/mysql/mod.rs
dlen: In the end I have created a PR on testcontainers-rs to support the addition of a health-cmd that would allow us to define a custom healthcheck and so not relying on container output or duration. If that gets merged latest changes in this branch will work fine (updating cargo version for testcontainers-rs). https://github.com/testcontainers/testcontainers-rs/pull/529

DDtKey: I'd not expect the PR to be reviewed & released quickly (last release has took ~ a year, due to lack of maintainers).
So would suggest to implement something working with current version and then we can update the implementation.

dlen: Alright! Sounds good. I have just pushed some changes to check on another condition on stderr. As the container always binds a port in the scenario of the testcontainers it makes sense. This might be good enough till a new testcontainers version. Let me know what you think and sorry for all the back and forth!

DDtKey: Thanks for all your effort to add one more image!

DDtKey:
Released as part of 0.1.3 version: https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.1.3

PR #66, opened 2023-11-14
feat: add Neo4j image

knutwalker:


mervyn-mccreight:
First of all thanks for the contribution!

To fix the failing PR title check you need to change the subject to start with lower-case (`Add neo4j image` vs. `add neo4j image`) I think.

knutwalker:
@mervyn-mccreight thanks! I updated the PR title, it should work now according to https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#subject-case. Can you run the check again in case it's not picked up by Github?

src/neo4j/mod.rs
```diff
@@ -0,0 +1,702 @@
...
+    /// Only a subset of Semantic Versions are supported.
+    /// The version must be of the format
+    ///
+    ///    MAJOR[.MINOR[.PATCH]]
+    ///
+    ///
+    /// # Errors
+    ///
+    /// If the version is not valid according to the format.
+    pub fn with_version(
```
mervyn-mccreight, line 142: I'm asking myself whether this check is really worth the additional dependency. If you provide a non existing tag the container startup will fail anyways.

Another question I ask myself is if we really want to force the user to use the official image :thinking:

@DDtKey WDYT?

src/neo4j/mod.rs
```diff
@@ -0,0 +1,702 @@
...
+        self.pass = Value::Value(pass.into());
+        self
+    }
+
+    /// Do not use any authentication on the testcontainer.
+    ///
+    /// Setting this will override any prior usages of [`Self::with_user`] and
+    /// [`Self::with_password`].
+    pub fn without_authentication(mut self) -> Self {
+        self.user = Value::Unset;
```
mervyn-mccreight, line 177: This could call with_user and with_password internally if I'm not mistaken.

Would make it easier to maintain IMO.

src/neo4j/mod.rs
```diff
@@ -0,0 +1,702 @@
...
+        self.plugins.extend_from_slice(plugins);
+        self
+    }
+
+    /// Create a new instance of a Neo4j image of the given version with the default user and password.
+    ///
+    /// # Panics
+    ///
+    /// If the version is not valid according to the format described in [`Self::with_version()`].
+    #[deprecated(since = "0.2.0", note = "Use `from_env().with_version()` instead.")]
```
mervyn-mccreight, line 238: Why do you want to include immediate deprecated functions? Is it for downwards compatibility for projects this image is already in use?

mervyn-mccreight (commented):
Thanks for the PR.

I had a look and primarily have left some questions that came to my mind.

src/neo4j/mod.rs
DDtKey: Agree, I think we should avoid version check. 
Users may want to use their own registries with custom tagging (not such a rare case)

src/neo4j/mod.rs
DDtKey: I believe all deprecated functions should be deleted, also not clear what does `0.2.0` mean here. Likely the version of another crate?

DDtKey (changes requested):
First of all, thank you for the contribution! 🚀 

It looks pretty nice and convenient to use, but I have couple of concerns:

- Ability to inspect user's ENV vars & files is questionable. 
  - any change of these variables would require major update (breaking change)
  - even if we want to have `from_env`, we need to think how to generalize this over all images, instead of implementing this to particular one (btw, could be a good outcome)
  - we should ensure that environment variables used only for `from_env` and it's consistent over time
- license check is also seems pretty hardcoded & depends on environment. See related comment


But mostly it's just topics to discuss, there seems to be no need for much effort to get it merged, just a cleanup

src/neo4j/mod.rs
mervyn-mccreight: Whoops. Missed that.

src/neo4j/mod.rs
mervyn-mccreight: I guess 0.2.0 would the next version of this crate considering merging this PR would include a new feature.

src/neo4j/mod.rs
knutwalker: Ok, good point. I'll remove the validation and the extra dependency.

src/neo4j/mod.rs
knutwalker: Ah my bad. I have the module as a separate crate in https://github.com/knutwalker/neo4j-testcontainers-rs and copied it from there. The deprecation note and versions refer to that crate. I'll remove the deprecated methods from this PR.

src/neo4j/mod.rs
```diff
@@ -0,0 +1,702 @@
...
+
+        Some(("NEO4JLABS_PLUGINS".to_owned(), plugin_definition))
+    }
+
+    fn conf_env(&self) -> impl IntoIterator<Item = (String, String)> {
+        let pass = Self::value(&self.pass)?;
+
+        if pass.len() < 8 {
+            Some((
+                "NEO4J_dbms_security_auth__minimum__password__length".to_owned(),
```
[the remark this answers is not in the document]
knutwalker: Yes, see https://neo4j.com/docs/operations-manual/current/docker/configuration/#docker-environment-variables for the Neo4j docs

src/neo4j/mod.rs
```diff
@@ -0,0 +1,702 @@
...
+    /// Return the connection URI to connect to the Neo4j server via Bolt over IPv4.
+    #[must_use]
+    pub fn bolt_uri_ipv4(&self) -> String {
+        let bolt_port = self
+            .state
+            .borrow()
+            .as_ref()
+            .expect("Container must be started before URI can be retrieved")
+            .host_port_ipv4(7687);
+        format!("bolt://127.0.0.1:{}", bolt_port)
```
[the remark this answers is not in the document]
knutwalker: I think it should be enough to expose just the port, I'll modify the methods accordingly.

src/neo4j/mod.rs
```diff
@@ -0,0 +1,702 @@
...
+        let acceptance_file = std::env::current_dir()
+            .ok()
+            .map(|o| o.join(ACCEPTANCE_FILE_NAME));
+
+        let has_license_acceptance = acceptance_file
+            .as_deref()
+            .and_then(|o| std::fs::File::open(o).ok())
+            .into_iter()
+            .flat_map(|o| std::io::BufReader::new(o).lines())
+            .any(|o| o.map_or(false, |line| line.trim() == image));
```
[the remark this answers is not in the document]
knutwalker: I was lifting this from the Java testcontainer [docs](https://java.testcontainers.org/modules/databases/neo4j/#choose-your-neo4j-license). It's [implemented](https://github.com/testcontainers/testcontainers-java/blob/1a3b23385f9a595576246a4c7f5dc54d244dd69d/modules/neo4j/src/main/java/org/testcontainers/containers/Neo4jContainer.java#L225) by using the [LicenseAcceptance](https://github.com/testcontainers/testcontainers-java/blob/1a3b23385f9a595576246a4c7f5dc54d244dd69d/core/src/main/java/org/testcontainers/utility/LicenseAcceptance.java#L18) utility, which seems to be from testcontainers itself, not something Neo4j specific.

I think this was also the reason to validate the version, so that users would have to use this method instead of just setting an enterprise version.

Since we removed the version validation, I can also remove this method here. I can still start an enterprise containers by using `RunnableImage` and that's good enough for us.

src/neo4j/mod.rs
```diff
@@ -0,0 +1,702 @@
...
+                pass.len().to_string(),
+            ))
+        } else {
+            None
+        }
+    }
+
+    fn build(mut self) -> Neo4jImage {
+        self.plugins.sort();
+        self.plugins.dedup();
```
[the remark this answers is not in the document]
knutwalker: IIRC I added the sorting to be better test it, since the list of plugins get translated into a JSON-ish env value. I switched to a BTreeSet instead.

src/neo4j/mod.rs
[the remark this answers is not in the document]
DDtKey: Thanks for the references!
We may think of implementing something similar actually - reusable util for license acceptance.

Let's start without this check and we may create a separate PR to support something like that, If it sounds good to you

knutwalker:
Thanks for the review and the comments!

>   * even if we want to have `from_env`, we need to think how to generalize this over all images, instead of implementing this to particular one (btw, could be a good outcome)
>   * we should ensure that environment variables used only for `from_env` and it's consistent over time

I did start building this testcontainers module for the development of neo4j-labs/neo4rs where I wanted to add integration tests. I didn't really draw a good line between things that are relevant for testing neo4rs in particular and what should generally be available on the testcontainer and I'd say all the env vars and `from_env` is part of that.

Perhaps a generalized env handling is a good idea, but then not introduced by this PR.

I removed all the env business and one this PR is merged and the module released, I can switch `neo4rs` over to use this and move all the env stuff to the `neo4rs` repo.

I also addressed all the other inline comments.

Looking forward to the next round 🏓

src/neo4j/mod.rs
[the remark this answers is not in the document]
knutwalker: Sure, sounds good. I already removed the method :)

src/neo4j/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+
+    /// Return the port to connect to the Neo4j server via HTTP over IPv6.
+    pub fn http_uri_ipv6(&self) -> u16 {
+        self.state
+            .borrow()
+            .as_ref()
+            .expect("Container must be started before port can be retrieved")
+            .host_port_ipv6(7474)
+    }
+}
```
DDtKey, line 231: I guess there is no `uri` methods anymore, so should we cleanup naming? 

Btw, is it possible to pass ENV parameter to change the ports? In that case it's also wrong to assume the ports always correct

If these ports can be changed by ENV variable passed to container, we should either:
- get ENV var to get correct port of container, otherwise do fallback to default
- remove these methods (but I like the idea to have methods to get specific ports without explicit mapping)

knutwalker: > I guess there is no `uri` methods anymore, so should we cleanup naming?
 

Oh no, I though I renamed all the URI references 🙈. Will update the names.

> Btw, is it possible to pass ENV parameter to change the ports? In that case it's also wrong to assume the ports always correct

Well, kinda, maybe…
On the one hand, you can configure any possible configuration of Neo4j by setting a certain ENV. On the other hand, you have to do that through `RunnableImage`, the plain `Neo4j(Image)` type does not allow you to change the port, no ENVs are picked up implicitly. I guess you also have to change the exposed ports since the Neo4j Docker image only exposes the default ports and you also cannot do that through the module type offered here. At this point, I'd say there is no accidental part left and you'd have to be trying quite hard to change the ports. You gonna have to familiar enough with testcontainers as well as Neo4j, that I would expect that you'd then also be able to figure out how the query for the mapped ports. So I guess, within reasonable effort, you cannot change the ports, though, _technically_ you can.

DDtKey: Ok, makes sense for now, we always can do a maintenance if anybody would ask

In general, even if you can override the port, it will be possible to `runnable_image..host_port_ipv6()` instead of `runnable_image.inner().bolt_port_ipv4()`. So these methods looks like a good defaults.

Thanks!

DDtKey (approved):
LGTM, thanks for all efforts!

Some notes after the PR:
- would be nice to have some reusable utility to accept license
- consider adding generalized `from_env` methods to our images

I'll prepare issues for these ones to cover & discuss

DDtKey:
@knutwalker new version has been released: https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.1.4

You can use it now from crates.io!

knutwalker:
@DDtKey Thanks for merging and releasing! I will update knutwalker/neo4j-testcontainers-rs to re-export this module and declare it EOL and will move the ENV and license handling to the neo4rs test setup.

PR #72, opened 2023-11-28
feat: add `Microsoft SQL Server` module

kymmt90:
This PR adds a Microsoft SQL Server module based on the official Docker images.

https://hub.docker.com/_/microsoft-mssql-server

Thanks!

### DIFF
--- a/examples/mssql_server.rs
+++ b/examples/mssql_server.rs
@@ -1,0 +1,31 @@
+use testcontainers_modules::{mssql_server::MssqlServer, testcontainers::clients::Cli};
+use tokio::net::TcpStream;
+use tokio_util::compat::TokioAsyncWriteCompatExt;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error + 'static>> {
+    let docker = Cli::default();
+    let image = MssqlServer::default();
+    let container = docker.run(image);
+
+    // Build Tiberius config
+    let mut config = tiberius::Config::new();
+    config.port(container.get_host_port_ipv4(1433));
+    config.authentication(tiberius::AuthMethod::sql_server(
+        "sa",
+        "yourStrong(!)Password",
+    ));
+    config.trust_cert();
+
+    // Connect to the database
+    let tcp = TcpStream::connect(config.get_addr()).await?;
+    tcp.set_nodelay(true)?;
+    let mut client = tiberius::Client::connect(config, tcp.compat_write()).await?;
+
+    // Run a test query
+    let stream = client.query("SELECT 1 + 1", &[]).await?;
+    let row = stream.into_row().await?.unwrap();
+    assert_eq!(row.get::<i32, _>(0).unwrap(), 2);
+
+    Ok(())
+}
--- a/src/mssql_server/mod.rs
+++ b/src/mssql_server/mod.rs
@@ -1,0 +1,162 @@
+use std::collections::HashMap;
+
+use testcontainers::{core::WaitFor, Image};
+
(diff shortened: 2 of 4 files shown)


src/mssql_server/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+/// The default value is `Developer`, which will run the container using the Developer Edition.
+#[derive(Debug)]
+pub struct MssqlServer {
+    env_vars: HashMap<String, String>,
+}
+
+impl MssqlServer {
+    const NAME: &'static str = "mcr.microsoft.com/mssql/server";
+    const TAG: &'static str = "2022-CU10-ubuntu-22.04";
+    const DEFAULT_SA_PASSWORD: &'static str = "yourStrong(!)Password";
```
DDtKey, line 62: If we gonna use custom password as default, I think we should either expose this constant to allow reusing or add `with_password` method directly to `MsSqlServer` to make it easier to configure, otherwise it will be always necessary to use `RunnableImage` to override the password or hardcode from the sources.

examples/mssql_server.rs
DDtKey, line 1: You need to add this example to `Cargo.toml`

DDtKey (commented):
Thanks for your contribution! 👍

src/mssql_server/mod.rs
kymmt90: Thanks for the comment. Added `MssqlServer::with_password` method. In addition to that, changed `SA_PASSWORD` to `MSSQL_SA_PASSWORD` because `SA_PASSWORD` seems to be deprecated.

https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-configure-environment-variables?view=sql-server-2017#environment-variables

> The SA_PASSWORD environment variable is deprecated. Use MSSQL_SA_PASSWORD instead.

kymmt90:
I'll take a look at the CI failures...

kymmt90:
The failed tests sometimes tried to login before [master db](https://learn.microsoft.com/en-us/sql/relational-databases/databases/master-database?view=sql-server-ver16) (a system db where SA user is in) was recovered and got auth errors.

```
 failures:
  
  ---- mssql_server::tests::one_plus_one stdout ----
  Error: Server(TokenError { code: 18456, state: 1, class: 14, message: "Login failed for user 'sa'.", server: "51a9d25accfc", procedure: "", line: 1 })
```

Fixed `ready_conditions` to be able to login as SA user sure. Now these tests are not flaky in my local environment.

@DDtKey Could you please run CI again for this changes?

src/mssql_server/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    fn tag(&self) -> String {
+        Self::TAG.to_owned()
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        // Wait until the server is ready for connections and all system databases are recovered
+        vec![WaitFor::StdOutMessage {
+            message: "Recovery is complete".to_owned(),
+        }]
```
DDtKey, line 102: One condition is enough? Shouldn't we have something like `ready to accept connections`?

kymmt90: Actually "SQL Server is now ready for client connections" (deleted the check in 1637d52) also appears before "Recovery is complete", it makes more sense to check both messages. I will update the condition.

DDtKey: Thanks a lot!

DDtKey:
Thanks for all efforts to get it merged!

New version has been released and available on crates.io
https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.2.0

kymmt90:
Thank you! 🥳

PR #77, opened 2023-12-29
feat: add `victoria_metrics` module

level-one-learner:


src/lib.rs
```diff
@@ -54,6 +54,9 @@ pub mod redis;
 #[cfg(feature = "trufflesuite_ganachecli")]
 #[cfg_attr(docsrs, doc(cfg(feature = "trufflesuite_ganachecli")))]
 pub mod trufflesuite_ganachecli;
+#[cfg(feature = "victoriametrics")]
+#[cfg_attr(docsrs, doc(cfg(feature = "victoriametrics")))]
```
DDtKey, line 58: What do you think of `victoria-metrics` as feature name?

DDtKey (commented):
Thanks for the contribution! 🚀 

One question and LGTM

src/lib.rs
level-one-learner: Sure, thank you. Do you know if I only need to change the two lines?

```rust
#[cfg(feature = "victoria-metrics")]
#[cfg_attr(docsrs, doc(cfg(feature = "victoria-metrics")))]
pub mod victoriametrics;
```

level-one-learner: @DDtKey I've updated the feature to use snake case everywhere.

DDtKey: Thank you!

DDtKey (approved):

PR #87, opened 2024-01-24
feat: add `SurrealDb` module

Jarsop:
This add `surrealdb` module. The version of `surrealdb` crate is fixed to a specific commit which bumps `rustls` version and enable compatibility with other crates. I will looking for the next release to switch on the tagged version.

And thanks for your work!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,3 +86,7 @@ required-features = ["neo4j"]
 [[example]]
 name = "mssql_server"
 required-features = ["mssql_server"]
+
+[[example]]
+name = "surrealdb"
+required-features = ["surrealdb"]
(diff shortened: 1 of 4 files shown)


DDtKey:

> The version of surrealdb crate is fixed to a specific commit which bumps rustls version and enable compatibility with other crates. I will looking for the next release to switch on the tagged version.

Since it's only for testing, I'm totally okay with that. We can wait, thanks for clarifying

DDtKey (approved):

Jarsop:
> Thank you for the contribution! 🚀
> 
> > The version of surrealdb crate is fixed to a specific commit which bumps rustls version and enable compatibility with other crates. I will looking for the next release to switch on the tagged version.
> 
> Since it's only for testing, I'm totally okay with that. We can wait, thanks for clarifying

You're welcome, I just pushed an example and one more test

DDtKey:
The module has been released: https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.3.2

Thank you!

PR #94, opened 2024-02-17
build(ci): enable `clean-per-run` for `cargo hack` in tests

DDtKey:
It makes CI more stable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,4 +114,4 @@ jobs:
         with:
           tool: cargo-hack
       - name: Tests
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -109,9 +109,8 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.toolchain }}
-      - uses: Swatinem/rust-cache@v2.7.3
```
DDtKey, line 112: It was duplicated

Issue #95, opened 2024-02-17
Redis Stack support


tsurai:
### Feature request description

I would like to add the `Redis Stack` image to use the various `JSON` commands in `Redis`. Would you rather have me add a second `RedisStack` struct in the same `redis` module or add it as a separate module?

### Image reference

`redis/redis-stack-server`

DDtKey:
Hi @tsurai!
I don't mind to add `Redis Stack` to existing `redis` module, but need to split them into two separate files. And in order not to cause breaking changes, we should re-export them through existing `mod.rs`

PR #98, opened 2024-02-18
feat: add `Redis Stack` module

tsurai:
This PR adds for `Redis Stack` with the tag `7.2.0-v8` to the already existing `redis` module as discussed in #95


src/redis/mod.rs
```diff
@@ -1,47 +1,5 @@
-use testcontainers::{core::WaitFor, Image};
+mod default;
```
DDtKey, line 1: let's use `standalone` instead of default? The module is private, so only for better understanding

src/redis/stack.rs
```diff
@@ -0,0 +1,51 @@
+use testcontainers::{core::WaitFor, Image};
+
+const NAME: &str = "redis/redis-stack-server";
+const TAG: &str = "7.2.0-v8";
+
+#[derive(Debug)]
+pub struct RedisStack;
```
DDtKey, line 7: Let's add documentation as for others, preferably for both `Redis` and `RedisStack`.
It was missing only because this module was moved from the old repository

DDtKey (commented):
Just couple of comments and LGTM!

src/redis/mod.rs
tsurai: I wasn't happy with default either but couldn't come up with an alternative in the moment. `standalone` is a lot better.

src/redis/mod.rs
```diff
@@ -1,47 +1,7 @@
...
-        let client = redis::Client::open(url.as_ref()).unwrap();
-        let mut con = client.get_connection().unwrap();
-
-        con.set::<_, _, ()>("my_key", 42).unwrap();
-        let result: i64 = con.get("my_key").unwrap();
-        assert_eq!(42, result);
-    }
-}
+pub use standalone::Redis;
+pub use stack::RedisStack;
```
DDtKey, line 7: Just to apply `rustfmt`
```suggestion
pub use stack::RedisStack;
pub use standalone::Redis;
```

src/redis/mod.rs
```diff
@@ -1,47 +1,7 @@
-use testcontainers::{core::WaitFor, Image};
+mod standalone;
+mod stack;
```
DDtKey, line 2:
```suggestion
mod stack;
mod standalone;
```

src/redis/mod.rs
DDtKey, line 7:
```suggestion
pub use stack::RedisStack;
pub use standalone::Redis;
```

DDtKey (commented):
`rustfmt` complains after renaming

DDtKey (approved):

DDtKey:
The module has been released https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.3.4

PR #109, opened 2024-03-30
feat: add cockroachdb image impl

jasondemorrow:
Adds Cockroach DB module


.gitignore
```diff
@@ -109,3 +109,6 @@ Temporary Items
 *.iml
 *.iws
 *.ipr
+
+# VSCode
+.vscode
```
DDtKey, line 114:
```suggestion
.vscode

```

src/cockroach_db/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct CockroachDbArgs;
+
+impl ImageArgs for CockroachDbArgs {
+    fn into_iterator(self) -> Box<dyn Iterator<Item = String>> {
+        Box::new(vec!["start-single-node".to_owned(), "--insecure".to_owned()].into_iter())
```
DDtKey, line 55: I guess these arguments can't be controlled with env-variables?

In that case I'd suggest to expose `insecure` as an option of arguments. To provide some control over the behavior

What alternatives are there instead of `start-single-node`?

DDtKey (commented):
Thanks for the contribution! 🙏 

Just one topic I want to discuss: image arguments

src/cockroach_db/mod.rs
jasondemorrow: Correct, those can't, as far I know or can find, be set via env vars. The documentation for cockroach command-line script can be found [here](https://www.cockroachlabs.com/docs/stable/cockroach-commands). The alternative to `start-single-node` is typically going to be `start`, which starts the node as part of a cluster. I think the common use-case for testcontainers is single-node, simple setup, which was why I hard-coded these. I can impl `Default` for these args perhaps, and expose a `new` that takes a `Vec<String>` in case someone wants to fire up a cluster with TLS enabled.

jasondemorrow:
Made suggested changes

src/cockroach_db/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+            tag,
+            env_vars: Default::default(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct CockroachDbArgs {
+    pub command: String,
+    pub args: Vec<String>,
```
DDtKey, line 57: I'd better encapsulate them. It will be easier to change without breaking changes.

For example, add `new` which accepts command with arguments and returns the instance. 

And inside we can even store only 1 `Vec`

jasondemorrow: I think encapsulating them is a good idea, but I prefer to keep the strings separate

jasondemorrow: Done

DDtKey (approved):
Thank you 🙏

src/cockroach_db/mod.rs
```diff
@@ -0,0 +1,127 @@
+use std::collections::BTreeMap;
+
+use testcontainers::{core::WaitFor, Image, ImageArgs};
+
+const DEFAULT_IMAGE_NAME: &str = "cockroachdb/cockroach";
+const DEFAULT_IMAGE_TAG: &str = "latest-v23.2";
```
mervyn-mccreight, line 6: I'd like to have stable tags (best effort) as a default for reproducibility reasons using this module.

My suggestion would be "v23.2.3" for now.

mervyn-mccreight:
Thanks for the contribution :)

I had a quick look and added one comment, open for discussion.

mervyn-mccreight (approved):

PR #111, opened 2024-04-01
feat: add `CNCF Distribution` module

esposm03:
This PR adds a module for CNCF Distrubution, a registry for container images. Some basic testing is included.

### DIFF
--- a/src/cncf_distribution/mod.rs
+++ b/src/cncf_distribution/mod.rs
@@ -1,0 +1,119 @@
+use testcontainers::{core::WaitFor, Image};
+
+const NAME: &str = "registry";
+const TAG: &str = "2";
+
+/// Module to work with a custom Docker registry inside of tests.
+///
+/// Starts an instance of [`CNCF Distribution`], an easy to use registry for container images.
+///
+/// # Example
+/// ```
+/// use testcontainers::clients;
+/// use testcontainers_modules::cncf_distribution;
+///
+/// let docker = clients::Cli::default();
+/// let registry = docker.run(cncf_distribution::CncfDistribution);
+///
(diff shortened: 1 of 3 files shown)


src/cncf_distribution/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn name(&self) -> String {
+        NAME.to_owned()
+    }
+
+    fn tag(&self) -> String {
+        TAG.to_owned()
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![WaitFor::millis(500)]
```
DDtKey, line 40: Is there any way to depend on stdout/stderr? It's much better, time is too flaky condition usually

DDtKey (commented):
Thanks for the contribution!

src/cncf_distribution/mod.rs
esposm03: Yes, I have just switched it from `millis` to `message_on_stderr`

src/cncf_distribution/Dockerfile.test
DDtKey, line 1: Sorry, missed this one

I'd prefer not to keep this file under the `src` directory 

In general it's so small that we can keep it hardcoded as constant string in tests

esposm03: Alright, I have just pushed a fix for this, by inlining the Dockerfile inside the source file.

mervyn-mccreight:
Can you resolve the existing conflicts please? I'd integrate this into the next release then :)

esposm03:
Done👍

mervyn-mccreight (approved):

PR #121, opened 2024-04-21
feat: add kwok module

Bakrog:
KWOK (Kubernetes WithOut Kubelet) is used to test pods and Kubernetes operators. It would be great if we could have it as an option here so that we can more easily use it on our Kubernetes operators tests.


src/kwok_cluster/mod.rs
```diff
@@ -0,0 +1,132 @@
+use testcontainers::{core::WaitFor, Image};
+
+const NAME: &str = "registry.k8s.io/kwok/cluster";
+const TAG: &str = "v0.4.0-k8s.v1.28.0";
```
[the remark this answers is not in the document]
mervyn-mccreight: Maybe it would be useful to upgrade this to v0.5.2 before releasing this module. I know it has been released quite recently, I just thought it might be a nice idea to start up to date :)

Cargo.toml
```diff
@@ -40,6 +40,7 @@ trufflesuite_ganachecli = []
 victoria_metrics = []
 zookeeper = []
 cockroach_db = []
+kwok_cluster = []
```
mervyn-mccreight, line 43: I think we should name the feature (and the module folder) `kwok`, because that's the official term as far as I understand.

mervyn-mccreight (changes requested):
Thanks for your contribution 🚀 .
Some minor comments from my side.

@DDtKey WDYT?

Cargo.toml
Bakrog: They have two container images. The cluster version (`registry.k8s.io/kwok/cluster:v0.5.2-k8s.v<k8s-version-here>`) is an `all-in-one` that runs a cluster without extra config. The other image (`registry.k8s.io/kwok/kwok:v0.5.2`) is basically to run inside a kubernetes cluster or pointing to an existing one.
I would suggest to keep the names similar with their releases so that others could distinguish them.

src/kwok_cluster/mod.rs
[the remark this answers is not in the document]
Bakrog: I pushed it :)

Cargo.toml
DDtKey: Actually, we can take a look at redis module:
it has standalone and cluster version, but they are under the same feature and module

So I'd suggest to use `kwok` feature here, but keep fully-qualified name of the image, e.g `KwokCluster`

Bakrog: Done.

DDtKey (approved):

src/kwok/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    use k8s_openapi::api::core::v1::Namespace;
+    use kube::{
+        api::ListParams,
+        client::Client,
+        config::{AuthInfo, Cluster, KubeConfigOptions, Kubeconfig, NamedAuthInfo, NamedCluster},
+        Api, Config,
+    };
+    use testcontainers::clients;
+
+    use crate::kwok_cluster::KwokCluster;
```
DDtKey, line 66: Could you fix import? Tests are failing

DDtKey:
We've released a new version of `testcontainers`: [0.16.0](https://github.com/testcontainers/testcontainers-rs/releases/tag/0.16.0) and [modules `.0.4.0`](https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.4.0) accordingly

Could you update tests to use new API? You can take a look at any module as an example 

Sorry and thank you in advance

Bakrog:
> We've released a new version of `testcontainers`: [0.16.0](https://github.com/testcontainers/testcontainers-rs/releases/tag/0.16.0) and [modules `.0.4.0`](https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.4.0) accordingly
> 
> Could you update tests to use new API? You can take a look at any module as an example
> 
> Sorry and thank you in advance

Hi, I rebased from main.

Bakrog:
> > We've released a new version of `testcontainers`: [0.16.0](https://github.com/testcontainers/testcontainers-rs/releases/tag/0.16.0) and [modules `.0.4.0`](https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.4.0) accordingly
> > Could you update tests to use new API? You can take a look at any module as an example
> > Sorry and thank you in advance
> 
> Hi, I pushed a rebase.

I'll push a fix for the test again, I still need to check what changed on the new version.

DDtKey (approved):

DDtKey:
Thank you!

PR #122, opened 2024-04-26
Upgrade Google Cloud SDK emulators image

psalaberria002:


DDtKey:
Hi @psalaberria002,

We usually don't update versions unless there are any changes to the image implementation.

You always can override the version with `RunnableImage::with_tag`. E.g:

```rs
let image = RunnableImage::from(MyImage).with_tag("1.1.1");

```

Thank you!

PR #123, opened 2024-04-27
feat: add nats module

haboumrad:
This PR adds a module for NATS messaging system


src/nats/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+        Self {
+            version: Cow::Borrowed(Self::DEFAULT_VERSION_TAG),
+            user: Some(Cow::Borrowed(Self::DEFAULT_USER)),
+            pass: Some(Cow::Borrowed(Self::DEFAULT_PASS)),
+        }
+    }
+
+    /// Set the Nats version to use.
+    /// The value must be an existing Nats version tag.
+    pub fn with_version(mut self, version: impl Into<Value>) -> Self {
```
DDtKey, line 34: We usually don't expose such methods as part of `Image` implementation. It's possible to override the tag with `RunnableImage`

haboumrad: [26daf99](https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/26daf99273e874cc6b792256e793301a0e56e198)

DDtKey:
We've released a new version of `testcontainers`: [0.16.0](https://github.com/testcontainers/testcontainers-rs/releases/tag/0.16.0) and [modules `.0.4.0`](https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.4.0) accordingly

Could you update tests to use new API? You can take a look at any module as an example

haboumrad:
> We've released a new version of `testcontainers`: [0.16.0](https://github.com/testcontainers/testcontainers-rs/releases/tag/0.16.0) and [modules `.0.4.0`](https://github.com/testcontainers/testcontainers-rs-modules-community/releases/tag/v0.4.0) accordingly
> 
> Could you update tests to use new API? You can take a look at any module as an example
> 
> Sorry and thank you in advance

No pb. Udate is done: https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/552289a24b42230b4bd49af5ae06c47f2c46ac0f

src/nats/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// The actual Nats testcontainers image type which is returned by `container.image()`
+pub struct NatsImage {
+    version: String,
+    auth: Option<(String, String)>,
+}
```
DDtKey, line 69: I think it's unnecessary complication to have a builder with `Cow` value wrappers. 

I'd prefer to provide a single struct `Nats`, which has builder-methods (accepts `impl Into<String>`) and implements `Image`.

Couple of reasons:
1. Keep as simple as possible
1.1 Only single public structure exposed, it provides everything we need
2. There is no real benefits, since we convert `Cow` to owned strings anyway. 
2.1. Empty string doesn't require allocation
2.2. Also, performance is not that important for testing, docker overhead is much more significant.
3. There is no need for explicit `impl From<Nats> for RunnableImage<NatsImage>`

So I suggest to have:

```rs
pub struct Nats {
    version: String,
    user: Option<String>,
    pass: Option<String>,
}

impl Nats {
 // all builder-methods and getters 
}

impl Default for Nats {..}

impl Image for Nats {..}

```

src/nats/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+            WaitFor::message_on_stderr("Server is ready"),
+        ]
+    }
+}
+
+impl Nats {
+    pub fn build(self) -> NatsImage {
+        let auth = self
+            .user
+            .and_then(|user| self.pass.map(|pass| (user.into_owned(), pass.into_owned())));
```
DDtKey, line 125: Btw, there is a shorter syntax for such operation: [zip method](https://doc.rust-lang.org/core/option/enum.Option.html#method.zip)

src/nats/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn tag(&self) -> String {
+        self.version.clone()
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![
+            WaitFor::message_on_stderr("Listening for client connections on 0.0.0.0:4222"),
+            WaitFor::message_on_stderr("Server is ready"),
+        ]
+    }
```
DDtKey, line 118: I don't really understand, how auth is configured? We have the parameters, but there is no any environment variables or comamnds

DDtKey (changes requested):
Thank you for your contribution, I've took a deeper look and left some comments

src/nats/mod.rs
haboumrad: you are correct. I took the neo4j integration as a template to follow. Simplification made done : [bca942f](https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/bca942f0af1f4f393a818ae132cc83c4b825415e)

src/nats/mod.rs
haboumrad: This method has been removed: [bca942f](https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/bca942f0af1f4f393a818ae132cc83c4b825415e)

src/nats/mod.rs
haboumrad: All authentication methods removed in the simplification: [bca942f](https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/bca942f0af1f4f393a818ae132cc83c4b825415e)

DDtKey: Wouldn't it be useful to provide the ability to specify authorization? I guess this should be an ENV variable

src/nats/mod.rs
DDtKey: Thank you!

Yeah, good point, need to review Neo4J in order to simplify. Thanks for pointing that out

src/nats/mod.rs
```diff
@@ -0,0 +1,63 @@
+use testcontainers::{core::WaitFor, Image};
+
+const NAME: &str = "nats";
+const TAG: &str = "2.10.14";
+
+/// Nats image for [testcontainers](https://crates.io/crates/testcontainers).
+///
+/// This image is based on the official [Nats](https://hub.docker.com/_/nats) image.
+#[derive(Debug, Default)]
+pub struct Nats;
```
DDtKey, line 10: Would be nice to make it not exhaustive in order to prevent backwards incompatible changes.

This way we won't break the API if we decide to add fields. `Nats::default().start()` will work, while `Nats.start()`

E.g:

```suggestion
pub struct Nats {
  _priv: (),
}
```

haboumrad: [67020e1](https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/67020e115076857b34d08546beed6aca3736464a)

src/nats/mod.rs
haboumrad: This is done through command line args` --user` and `--pass`.

Still, I am not sure to understand in which use case it would be useful to add customization for credentials in integration test.

haboumrad: https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/82d0fae11585c21f478281fc38e705b29c1bec59

src/nats/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+/// This image is based on the official [Nats](https://hub.docker.com/_/nats) image.
+#[derive(Debug, Default)]
+pub struct Nats {
+    _private: (),
+}
+
+#[derive(Debug, Clone)]
+pub struct NatsServerArgs {
+    pub user: Option<String>,
+    pub pass: Option<String>,
```
DDtKey, line 17: I think the last request:

Let's keep them private and expose builder methods: `with_user` and `with_pass`

It also will help to avoid breaking changes

haboumrad: https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/5042849ea41a48c687b4e0d089ebe84e435457ae

DDtKey (approved):

src/nats/mod.rs
haboumrad: https://github.com/testcontainers/testcontainers-rs-modules-community/pull/123/commits/5042849ea41a48c687b4e0d089ebe84e435457ae

DDtKey (approved):

PR #129, opened 2024-05-05
feat: k3s module

dghilardi:
Add k3s module - could be useful for tests that needs to run inside a kubernetes cluster


src/k3s/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+pub const RANCHER_WEBHOOK_PORT: u16 = 8443;
+
+#[derive(Debug, Default, Clone)]
+pub struct K3s {
+    env_vars: HashMap<String, String>,
+    conf_mount: Option<Mount>,
+}
+
+#[derive(Default, Debug, Clone)]
+pub struct K3sArgs;
```
DDtKey, line 22: I'd suggest to make it not-exhaustive, to be able to extend in the future without breaking changes.
E.g:
```rust
pub struct K3sArgs {
    _priv: (),
}

```

src/k3s/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl ImageArgs for K3sArgs {
+    fn into_iterator(self) -> Box<dyn Iterator<Item = String>> {
+        Box::new([String::from("server"), String::from("--snapshotter=native")].into_iter())
```
DDtKey, line 26: Don't we need to expose an ability to specify/override `snapshotter`?

src/k3s/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+use testcontainers::{Image, ImageArgs};
+
+const NAME: &str = "rancher/k3s";
+const TAG: &str = "v1.28.8-k3s1";
+pub const TRAEFIK_HTTP: u16 = 80;
+pub const KUBE_SECURE_PORT: u16 = 6443;
+pub const RANCHER_WEBHOOK_PORT: u16 = 8443;
+
+#[derive(Debug, Default, Clone)]
+pub struct K3s {
```
DDtKey, line 16: Could you add a simple documentation with references to k3s and docker image? Example: https://github.com/testcontainers/testcontainers-rs-modules-community/blob/06efbb0474d1718efa459c718eff885e9354e833/src/rabbitmq/mod.rs#L6-L28

DDtKey:
Thanks for the contribution 🙏 

I've left couple of comments

src/k3s/mod.rs
dghilardi: I added the _priv field to make the struct non-exaustive

src/k3s/mod.rs
dghilardi: Sure, I changed to keep native as default (I believe it is the most useful for test purposes) but now it can be customized via args field.

src/k3s/mod.rs
dghilardi: I added a bit of documentation and an example like the rabbitmq module you linked

src/k3s/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+/// [`K3s docker image`]: https://hub.docker.com/r/rancher/k3s
+#[derive(Debug, Default, Clone)]
+pub struct K3s {
+    env_vars: HashMap<String, String>,
+    conf_mount: Option<Mount>,
+}
+
+#[derive(Debug, Clone)]
+pub struct K3sArgs {
+    pub snapshotter: Option<String>,
```
DDtKey, line 50: Let's make the field private, and provide public builder-like method `with_snapshotter(value: impl Into<String>)`. And LGTM!

Thus, we can remove `_priv` and the usage will be like that:

```rs
K3sArgs::default().with_snapshotter("..");
```

Pros of this:
- encapsulated field (=> we can change underlying storage format for arguments if we wish)
- allows to pass not only string

DDtKey (approved):

src/k3s/mod.rs
dghilardi: Ok, I did as you suggested :slightly_smiling_face:

DDtKey (approved):

PR #138, opened 2024-05-21
chore(deps): update `dynamodb-local` default image tag 

quettabit:
Recently, DynamoDB local added support for two more DynamoDB API features.
https://aws.amazon.com/about-aws/whats-new/2023/12/amazon-dynamodb-local-two-api-features/

### DIFF
--- a/src/dynamodb_local/mod.rs
+++ b/src/dynamodb_local/mod.rs
@@ -1,7 +1,7 @@
 use testcontainers::{core::WaitFor, Image};
 
 const NAME: &str = "amazon/dynamodb-local";
-const TAG: &str = "2.0.0";
+const TAG: &str = "2.2.0";
 const DEFAULT_WAIT: u64 = 3000;
 
 #[derive(Default, Debug)]


quettabit:
hmm, given there is support for using images with custom tags, i'm wondering when should one update the tags like this in the module.

DDtKey:
We should avoid updating the default version **as long as the image can work by simply overriding the version**.

Because changing of default version may lead to changed behavior of the module and broke someone's build/test suite. And should be considered as breaking change of module (not all images follow semver, or can just violate it by accident)

I'd recommend to stick to it in your code with [`RunnableImage::with_tag`](https://docs.rs/testcontainers/latest/testcontainers/core/struct.RunnableImage.html#method.with_tag)

quettabit:
hey @DDtKey! no worries, totally makes sense. (btw, thank you for all the efforts around `testcontainers-rs` and `testcontainers-rs-modules-community`. i'm coming back to rust testcontainers after ~3yrs, and things seem much nicer now.)

PR #143, opened 2024-06-09
feat!: add mongodb replset option with example

sinyo-matu:
Thank you for the amazing work.
I really enjoy using testcontainers with Rust.

When working with MongoDB, it is very common to need to test [transaction operations](https://www.mongodb.com/docs/manual/core/transactions/). Therefore, I added some options to facilitate this.

- Added a new struct MongoReplSet, which represents a MongoDB server started with [mongod --replSet](https://www.mongodb.com/docs/manual/reference/program/mongod/#std-option-mongod.--replSet). This enables Replica Set which is necessary for the transaction operations
- Added with_name and with_tag for users' convenience to choose their preferred image and tag.
- Added some relevant tests.
- Breaking Change: `Mongo.start()` will no longer work. Instead, users need to explicitly call `Mongo::default()`.


src/mongo/standalong.rs
```diff
@@ -0,0 +1,105 @@
...
+impl Mongo {
+    pub fn with_name(mut self, name: impl AsRef<str>) -> Self {
+        self.name = name.as_ref().to_string();
+        self
+    }
+
+    pub fn with_tag(mut self, tag: impl AsRef<str>) -> Self {
+        self.tag = tag.as_ref().to_string();
+        self
+    }
```
DDtKey, line 29: We usually don't expose these methods because it's possible to override with [RunnableImage::with_name](https://docs.rs/testcontainers/latest/testcontainers/core/struct.RunnableImage.html#method.with_name) (and `with_tag`)

src/mongo/mod.rs
```diff
@@ -1,61 +1,17 @@
...
-const NAME: &str = "mongo";
-const TAG: &str = "5.0.6";
-
-#[derive(Default, Debug)]
-pub struct Mongo;
-
-impl Image for Mongo {
-    type Args = ();
+pub use replset::MongoReplSet;
+pub use standalong::Mongo;
```
DDtKey, line 5: I guess it should be `standalone`?

src/mongo/replset.rs
DDtKey, line 1: As far as I can see, the main difference is `ReplSetArgs` and `exec_after_start`. Right?

I think it can be simplified to single image with different options to start 🤔 

For example:
- expose only single `Mongo` image
- add `Mongo::standalone()` (or just `new` as default one) which doesn't have arguments
- add `Mongo::repl_set()` - which sets the necessary arguments
- `exec_after_start` depends on created kind of instance


I think the issue is how arguments are represented in current version of `testcontainers`. I'd like to refactor this part, so image will have override-able method `args` (just like `env_vars` and etc).
I'll prepare refactoring PR to `testcontainers`, but it might be released with some delay

DDtKey (changes requested):
Thank you for the contribution! 🙏  I really appreciate it!

I left couple of comments, please take a look and let me know what do you think.

Also, since we're touching the module, it would be great to add some basic documentation, for example like this one: https://github.com/testcontainers/testcontainers-rs-modules-community/blob/edf3c6e940521be56edb17b4c74477f5b6a1fada/src/rabbitmq/mod.rs#L6-L27

sinyo-matu:
@DDtKey 
Thanks for your kind comments!

> I think the issue is how arguments are represented in current version of testcontainers. I'd like to refactor this part, so image will have override-able method args (just like env_vars and etc).
>I'll prepare refactoring PR to testcontainers, but it might be released with some delay

I totally agree with you that we should expose only a single Mongo image for the user's convenience. As you mentioned, the representation of arguments makes it difficult to run different arguments driven by different options. Refactoring will be really helpful!

I think I may refactor my code after the release of testcontainers that includes the PR.

sinyo-matu:
And I would like to add  examples for mongodb

DDtKey:
Hi @sinyo-matu 👋 

Sorry for delay, totally forgot to notify you that a new version of testcontainers has been released.

Now it's possible to encapsulate arguments as part of image itself!

src/mongo/replset.rs
sinyo-matu: @DDtKey 
I updated my code. Please have a look.

sinyo-matu:
Sorry for the delay!
- Update my code. it is now following the new `Image` trait.
- Upgrade `mongodb` to 3.0.1.
- Add a simple example for mongo module

src/mongo/mod.rs
```diff
@@ -18,6 +48,35 @@ impl Image for Mongo {
...
+                "--eval".to_string(),
+                "'rs.initiate()'".to_string(),
+            ])
+            .with_cmd_ready_condition(CmdWaitFor::message_on_stdout(
+                "Using a default configuration for the set",
+            ))
+            // Wait for the replica set to be ready
+            // This is a workaround because the replica set isn't immediately ready
+            // Without this delay, an immediate connection to the replica set would fail
+            .with_container_ready_conditions(vec![WaitFor::seconds(2)])]),
```
DDtKey, line 77: Is there something better than hardcoded delay? (log, http call?)
Just wondering if we can make it more reliable

DDtKey (commented):
Thank you!
Just one question, but in general looks like LGTM

src/mongo/mod.rs
sinyo-matu: You're right. Let me do some investing

sinyo-matu: @DDtKey 
updated!

sinyo-matu: And could you tell me how I should fix my PR name?

src/mongo/mod.rs
```diff
@@ -18,6 +48,39 @@ impl Image for Mongo {
...
+                ])
+                .with_cmd_ready_condition(CmdWaitFor::message_on_stdout(
+                    "Using a default configuration for the set",
+                )),
+                ExecCommand::new(vec![
+                    "mongosh".to_string(),
+                    "--eval".to_string(),
+                    "'rs.status()'".to_string(),
+                ])
+                .with_cmd_ready_condition(CmdWaitFor::message_on_stdout("ok: 1")),
```
DDtKey, line 80: I think `rs.status` doesn't wait for "ok", right?

In that case we probably need to wrap it into something like:

```
until mongosh --eval 'rs.status()'; do sleep 1; done
```

WDYT?

src/mongo/mod.rs
DDtKey: I think there was a missing space, I've changed the naming a bit

DDtKey (approved):

sinyo-matu:
@DDtKey 
Thank you for the code review Artem

PR #148, opened 2024-06-16
feat: add Hashicorp Vault test container and test

jasondemorrow:
also adds tracing library to dev deps b/c I wanted to see the TC logs and figured others might too


jasondemorrow:
Tests in untouched code are failing, perhaps due to added tracing dep? That doesn't make sense, but it's all I can think of.

jasondemorrow:
Overrode my local 1.79 with 1.75 to match CI, but now aws-sqs won't build b/c it needs 1.76+?

jasondemorrow:
I see what happened here. I had the distinct misfortune of issuing a PR just as the upgrade to TC 0.18 landed in main :joy:

DDtKey (approved):
Thank you for the contribution 🚀

And sorry for this inconvenience after `testcontainers` update.

PR #154, opened 2024-06-29
feat: add a `Meilisearch` module

CommanderStorm:
Hi,

This PR adds support for [meilisearch](https://www.meilisearch.com/).

I have not [included all instance options](https://www.meilisearch.com/docs/learn/configuration/instance_options), and have limited this to the ones which are likely benefitial to testing. (having a second pair of eyes look over the list is likely benefitial)

Should I add an example here? https://github.com/testcontainers/testcontainers-rs-modules-community/tree/main/examples

I am a somewhat new Rust developer. I have stuck to the standards set out in the `reddis` and `postgres` containers.
If there is anything that you would like changed, I would appreciate the feedback ^^

PS: Great work on the library. It is a joy to use ❤️


src/meilisearch/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+///     "http://{}:{}",
+///     meilisearch_instance.get_host().unwrap(),
+///     meilisearch_instance.get_host_port_ipv4(7700).unwrap()
+/// );
+/// ```
+///
+/// [`Meilisearch`]: https://www.meilisearch.com/
+/// [`Meilisearch docker docs`]: https://www.meilisearch.com/docs/guides/misc/docker
+/// [`Meilisearch docker image`]: https://hub.docker.com/_/getmeili/meilisearch
+#[derive(Debug)]
```
CommanderStorm, line 33: Some derive `Clone` here. Should this be added?

```suggestion
#[derive(Debug, Clone)]
```

DDtKey: It doesn’t bring any issues for us, but I think it might be useful in some cases, so let’s add it. 

Thank you!

DDtKey (commented):
Thank you for the contribution !🙏

It looks good to me, let’s apply cargo fmt and LGTM 🚀

> I have not [included all instance options](https://www.meilisearch.com/docs/learn/configuration/instance_options), and have limited this to the ones which are likely benefitial to testing. (having a second pair of eyes look over the list is likely benefitial)

That’s totally fine, we shouldn’t provide everything. Too specific options always can be enabled/overridden by user 



> Should I add an example here? https://github.com/testcontainers/testcontainers-rs-modules-community/tree/main/examples

That’s totally up to you, not required, but something nice to have 🙂

src/meilisearch/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+        } else {
+            self.env_vars
+                .insert("MEILI_NO_ANALYTICS".to_owned(), "true".to_owned());
+        }
+        self
+    }
+
+    /// Sets the environment of the [`Meilisearch`] instance.
+    /// Default is [Environment::Development] if not overridden by this function.
+    /// Setting it to [Environment::Production] requires authentication via [Meilisearch::with_master_key]
```
DDtKey, line 82: We could extend the enum to require the key, like `Environment::Production(key)`, but it might be confusing that there are two ways to set master key though. However I guess it’s just impossible to start production without the key.

Just a note

CommanderStorm: I thought about this.
The reason why I did model it the way I did is that the master key can be set without setting the environment to prod..

(I can change it if you prefer this changed)

DDtKey: No-no, that makes sense actually

DDtKey (approved):

PR #156, opened 2024-06-30
chore: made deriving `Clone` consistent

CommanderStorm:
This is a followup to https://github.com/testcontainers/testcontainers-rs-modules-community/pull/154#discussion_r1660223826

(I don't have a real usecase for this. Just thought that more consistency can't hurt)

`Neo4jImage` is ommited, as `ContainerState` does not implement `Clone`.

### DIFF
--- a/src/elastic_search/mod.rs
+++ b/src/elastic_search/mod.rs
@@ -8,7 +8,7 @@ use testcontainers::{
 const NAME: &str = "docker.elastic.co/elasticsearch/elasticsearch";
 const TAG: &str = "7.16.1";
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
 pub struct ElasticSearch {
     _priv: (),
 }
--- a/src/clickhouse/mod.rs
+++ b/src/clickhouse/mod.rs
@@ -26,7 +26,7 @@ const CLICKHOUSE_PORT: ContainerPort = ContainerPort::Tcp(8123);
 ///
 /// [`ClickHouse`]: https://clickhouse.com/
 /// [`Clickhouse docker image`]: https://hub.docker.com/r/clickhouse/clickhouse-server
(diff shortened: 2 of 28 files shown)


src/elastic_search/mod.rs
```diff
@@ -8,7 +8,7 @@ use testcontainers::{
 const NAME: &str = "docker.elastic.co/elasticsearch/elasticsearch";
 const TAG: &str = "7.16.1";
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
 pub struct ElasticSearch {
     _priv: (),
```
CommanderStorm, line 13: One thing that I found interesting is this pattern vs (for example) `pub struct DynamoDb;`

Out of curiosity:
Is there a reason for having this unused member variable? ^^

DDtKey: >  Is there a reason for having this unused member variable? ^^

Yes, to provide encapsulation, this allows the image to be extended without breaking future changes.

`pub struct DynamoDb;` allows you to create an instance simply by using `DynamoDb`. Therefore adding a new field will break this usage.
Although `ElasticSearch` requires the use of `new()` or `default()` so that we can safely add new fields.

However, some images are still missing this pattern and should be updated in the future

DDtKey (approved):
Thank you very much!

I'd consider this as a feature, we introduce new allowed functionality in general.

> (I don't have a real usecase for this. Just thought that more consistency can't hurt)

I think one possible use-case is to spawn several configured instances, something like:

```rs
let image = MssqlServer::default().with_sa_password("my_password");
let container_1 = image.clone().start();
let container_2 = image.start()
```

It's unlikely to be common pattern, but there is no reason to restrict users of doing this.

DDtKey:
Forgot to change `chore` to `feat`, but not that important

PR #178, opened 2024-08-12
feat: add support for `apache\kafka` docker image

milenkovicm:
Up till now kafka test container used `confluentinc/cp-kafka`, recently kafka started producing two official
images `apache\kafka` and `apache\kafka-native` which may use `kraft` protocol, thus no zookeeper
instance is needed. 

`apache\kafka-native` image has kafka server compiled using GraalVM which decreases container startup and memory consumption.

Default image is `apache\kafka-native` which provide fast starting speed.

More details:

- [KIP-974: Docker Image for GraalVM based Native Kafka Broker](https://cwiki.apache.org/confluence/display/KAFKA/KIP-974%3A+Docker+Image+for+GraalVM+based+Native+Kafka+Broker)
- [KIP-975: Docker Image for Apache Kafka](https://cwiki.apache.org/confluence/display/KAFKA/KIP-975%3A+Docker+Image+for+Apache+Kafka)
- [Java testcontainer implementation](https://github.com/testcontainers/testcontainers-java/blob/1.20.1/modules/kafka/src/main/java/org/testcontainers/kafka/KafkaContainer.java#)


src/lib.rs
```diff
@@ -39,6 +39,9 @@ pub mod k3s;
 #[cfg(feature = "kafka")]
 #[cfg_attr(docsrs, doc(cfg(feature = "kafka")))]
 pub mod kafka;
+#[cfg(feature = "kafka_native")]
+#[cfg_attr(docsrs, doc(cfg(feature = "kafka_native")))]
+pub mod kafka_native;
```
DDtKey, line 44: I'd suggest to keep it under `kafka` feature. We don't introduce any dev-time deps or something too specific.

WDYT of `kafka/native.rs`?

src/kafka_native/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+const START_SCRIPT: &str = "/opt/kafka/testcontainers_start.sh";
+const DEFAULT_INTERNAL_TOPIC_RF: usize = 1;
+const DEFAULT_CLUSTER_ID: &str = "5L6g3nShT-eMCtK--X86sw";
+const DEFAULT_BROKER_ID: usize = 1;
+
+#[derive(Debug, Clone)]
+pub struct Kafka {
+    env_vars: HashMap<String, String>,
+    image_name: String,
+}
```
DDtKey, line 22: Let's add some documentation, like this one: https://github.com/testcontainers/testcontainers-rs-modules-community/blob/71875a6483e43e4ff953e42c26e08825d2f7f0f3/src/rabbitmq/mod.rs#L6-L27

Including references to images

src/lib.rs
milenkovicm: It definitely makes sense to go under kafka, question is should the `confluent` image be phased out at some point? 
If it will be phased out, your suggestion would be non breaking change. 

I'd also suggest to name it `kafka/apache.rs`. 
Only thing missing is how would users be able to change between apache and confluent version, WDYT? any suggestions?

DDtKey: I think we can put them into:
- `kafka/apache`
- `kafka/confluent`

And re-export module from `confluent` within `kafka` (with a comment about backward compatibility)
I think we may decide to keep both at the end and only breaking change in the future would be removal of re-exports.

Users can just import desired one.

milenkovicm:
I've used this commit to fix mongo example missing feature which was failing tests

src/kafka/apache.rs
```diff
@@ -0,0 +1,314 @@
...
+/// Module to work with [`Apache Kafka`] broker
+///
+/// Starts an instance of Apache Kafka broker, with Apache Kafka Raft (KRaft) is the consensus protocol
+/// enabled.
+///
+/// This module is based on the official [`Apache Kafka docker image`](https://hub.docker.com/r/apache/kafka)
+///
+/// Module comes in two flavours:
+///
+/// - [`Apache Kafka GraalVM docker image`](https://hub.docker.com/r/apache/kafka), which is default as it provides faster startup and lower memory consumption.
```
DDtKey, line 27: should it be kafka-native reference?

DDtKey (approved):
Thank you for the contribution 🙏

src/kafka/apache.rs
milenkovicm: Yes, it should, my bad

src/kafka/apache.rs
```diff
@@ -0,0 +1,321 @@
...
+    pub fn with_jvm_image(mut self) -> Self {
+        self.image_name = KAFKA_IMAGE_NAME.to_string();
+
+        self
+    }
+    /// Specify image tag
+    pub fn with_tag(mut self, tag: &str) -> Self {
+        self.tag = tag.to_string();
+        self
+    }
```
DDtKey, line 108: Sorry, one more comment since it's been changed:  we don't usually expose methods that are unified across all images.

It's possible to use [ImageExt::with_tag](https://docs.rs/testcontainers/latest/testcontainers/core/trait.ImageExt.html#tymethod.with_tag) to override the tag

milenkovicm: all right, will revert commit

milenkovicm: done

DDtKey (approved):

CommanderStorm (changes requested):
Hi, I left a few minor comments below, mostly about documentation ^^

src/kafka/apache.rs
```diff
@@ -0,0 +1,314 @@
...
+    fn name(&self) -> &str {
+        self.image_name.as_str()
+    }
+
+    fn tag(&self) -> &str {
+        TAG
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        vec![]
```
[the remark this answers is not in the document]
milenkovicm: yeah, it looks strange but container will not emit anything as it is in a while loop, waiting for starting script creation in the `exec_after_start`.
`exec_after_start` has a wait contidion on kafka to be started.

milenkovicm:
thanks @CommanderStorm, I hope your comments have been addressed

src/kafka/apache.rs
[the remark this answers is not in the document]
CommanderStorm: Could you add a comment stating this (for future devs)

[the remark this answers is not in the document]
milenkovicm: sure, no problem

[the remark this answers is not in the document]
milenkovicm: should be resolved now

CommanderStorm (approved):
LGTM, thanks for the changes

src/kafka/apache.rs
```diff
@@ -0,0 +1,335 @@
+use std::{borrow::Cow, collections::HashMap};
+use testcontainers::{
+    core::{ContainerPort, ContainerState, ExecCommand, WaitFor},
+    Image,
+};
+
+const KAFKA_NATIVE_IMAGE_NAME: &str = "apache/kafka-native";
+const KAFKA_IMAGE_NAME: &str = "apache/kafka";
+const TAG: &str = "latest";
```
DDtKey, line 9: Oops, missed that part.
We do not use the latest version because it does not guarantee image stability. Users can always override it to the latest version if they want, but this easily leads to instability and the implementation can quickly become outdated.

DDtKey, line 9: Oops, missed that part.
We do not use the latest version because it does not guarantee image stability. Users can always override it to the latest version if they want, but this easily leads to instability and the implementation can quickly become outdated.

DDtKey (changes requested):

src/kafka/apache.rs
DDtKey, line 9:
```suggestion
const TAG: &str = "3.8.0";
```

DDtKey (approved):

PR #180, opened 2024-08-19
feat: added an `OpenLDAP` testcontainer module

CommanderStorm:
I noticed that there currently is not LDAP testcontainer.

Similar to the [go testontainer modules](https://github.com/testcontainers/testcontainers-go/blob/c44b1b2ed935455fa222f122a7474efe309f641c/modules/openldap/openldap.go#L128C19-L128C41), I chose the [`bitnami/openldap`](https://github.com/bitnami/containers/blob/main/bitnami/openldap/README.md) containers.

Similar to https://github.com/testcontainers/testcontainers-rs-modules-community/pull/178, this PR includes the mongodb-example-fix.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ nats = []
 neo4j = []
 oracle = []
 orientdb = []
+openldap = []
 parity = []
 postgres = []
 rabbitmq = []
@@ -66,6 +67,7 @@ aws-types = "1.0.1"
 bollard = "0.17.0"
 futures = "0.3"
 lapin = "2.3.1"
+ldap3 = "0.11.5"
 meilisearch-sdk = "0.26.1"
 mongodb = "3.0.1"
 mysql = "25.0.0"
@@ -116,3 +118,11 @@ required-features = ["mssql_server"]
 [[example]]
 name = "surrealdb"
 required-features = ["surrealdb"]
+
+[[example]]
+name = "openldap"
+required-features = ["openldap"]
+
+[[example]]
+name = "mongo"
+required-features = ["mongo"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,8 @@ cockroach_db = []
 kwok = []
 
 [dependencies]
-parse-display = { version = "0.10.0", optional = true, default-features = false, features = [] }
+# TODO: update parse-display after MSRV>=1.80.0 bump of `testcontainer-rs` and `testcontainers-modules`
+parse-display = { version = "0.9.1", optional = true, default-features = false, features = [] }
 testcontainers = { version = "0.21.0" }
 
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ nats = []
 neo4j = []
 oracle = []
 orientdb = []
-openldap = []
+openldap = ["dep:parse-display"]
 parity = []
 postgres = []
 rabbitmq = []
@@ -54,6 +54,7 @@ cockroach_db = []
 kwok = []
 
 [dependencies]
+parse-display = { version = "0.10.0", optional = true, default-features = false, features = [] }
 testcontainers = { version = "0.21.0" }
 
 
--- a/src/openldap/mod.rs
+++ b/src/openldap/mod.rs
@@ -225,23 +225,15 @@ impl OpenLDAP {
     /// Set hash to be used in generation of user passwords.
     /// Default: [`PasswordHash::SSHA`].
     pub fn with_ldap_password_hash(mut self, password_hash: PasswordHash) -> Self {
-        self.env_vars.insert(
-            "LDAP_PASSWORD_HASH".to_owned(),
-            match password_hash {
-                PasswordHash::SSHA => "{SSHA}".to_owned(),
-                PasswordHash::SHA => "{SHA}".to_owned(),
-                PasswordHash::SMD5 => "{SMD5}".to_owned(),
-                PasswordHash::MD5 => "{MD5}".to_owned(),
-                PasswordHash::CRYPT => "{CRYPT}".to_owned(),
-                PasswordHash::CLEARTEXT => "{CLEARTEXT}".to_owned(),
-            },
-        );
+        self.env_vars
+            .insert("LDAP_PASSWORD_HASH".to_owned(), password_hash.to_string());
         self
     }
 }
 
 /// hash to be used in generation of user passwords.
-#[derive(Default, Debug, Clone, Copy, Eq, PartialEq)]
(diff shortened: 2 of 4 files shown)


CommanderStorm:
The testcase failour seems to be because https://github.com/paupino/rust-decimal/issues/669
We could solve this by adding ` --locked` to the `cargo-hack` command.
With that flag, the PR tests were to be green

src/openldap/mod.rs
```diff
@@ -0,0 +1,519 @@
...
+            password: password.to_owned(),
+        });
+        self
+    }
+
+    /// Add users (can be called multiple times)
+    /// Default: `[]`
+    ///
+    /// Alternatively, users can be added via [`OpenLDAP::with_user`].
+    pub fn with_users<Username: AsRef<str>, Password: AsRef<str>>(
```
DDtKey, line 160: We actually convert it to `String` anyway. So, I’d prefer to expect `impl Into<String>` thus we can avoid extra allocation if `String` has been passed to functions. And still works for `&str`.

Could you re-check all functions?
Thank you 🙏

src/openldap/mod.rs
```diff
@@ -0,0 +1,519 @@
...
+    fn tag(&self) -> &str {
+        TAG
+    }
+
+    fn ready_conditions(&self) -> Vec<WaitFor> {
+        // maybe OpenLDAP will have a healthcheck someday
+        // https://github.com/osixia/docker-openldap/issues/637
+        vec![
+            WaitFor::message_on_stderr("** Starting slapd **"),
+            WaitFor::seconds(2), // ideally, one should wait for a port instead of waiting a fixed time
```
DDtKey, line 361: Probably we can utilize HTTP wait strategy? 🤔

DDtKey (commented):
Thank you for the contribution 🚀 

I’ve left couple of comments, but looks good overall 👍

src/openldap/mod.rs
CommanderStorm: good call, have changed as such. Makes the code even a bit tidier (even though that part is subjective ^^)

src/openldap/mod.rs
```diff
@@ -0,0 +1,519 @@
...
+            log_success,
+            log_purge,
+            log_old,
+            log_old_attribute,
+        }: AccesslogSettings,
+    ) -> Self {
+        self.env_vars
+            .insert("LDAP_ENABLE_ACCESSLOG".to_owned(), "yes".to_owned());
+        let log_operations = match log_operations {
+            AccesslogLogOperations::Writes => "writes".to_owned(),
```
DDtKey, line 100: "writes" in lowercase, but everything else is in capital letters. Is it expected? (case insensitive?)

Let's probably utilize [parse-display](https://crates.io/crates/parse-display) for the enum. This will reduce boilerplate and should be less error prone

src/openldap/mod.rs
DDtKey: I see, thank you

Yes, let's use delay for now and later we can add the strategy

src/openldap/mod.rs
CommanderStorm: Okay, This is hitting my rust-knowlege-limits.

I need an `[String]`, to push into the vector
=> the best I can make the borrow-checker happy with is `<S: Into<String> + Clone>`.. Seems very unergonomic..

Should I switch back to just using `String`?

src/openldap/mod.rs
CommanderStorm: Good catch!
Everything here should be lowercase, will refactor using that dependency

src/openldap/mod.rs
```diff
@@ -0,0 +1,517 @@
...
+        user_password: &[(Username, Password)],
+    ) -> Self {
+        for (username, password) in user_password {
+            self.users.push(User {
+                username: username.clone().into(),
+                password: password.clone().into(),
+            })
+        }
+        self
+    }
```
DDtKey, line 168: Sorry, just realized that it's a slice of tuples. Thus, we can't avoid allocation
In that case, `ToString` may be better than `Into<String> + Clone`

```suggestion
    pub fn with_users<Username: ToString, Password: ToString>(
        mut self,
        user_password: &[(Username, Password)],
    ) -> Self {
        for (username, password) in user_password {
            self.users.push(User {
                username: username.to_string(),
                password: password.to_string(),
            })
        }
        self
    }
```

Cargo.toml
```diff
@@ -53,6 +54,7 @@ cockroach_db = []
 kwok = []
 
 [dependencies]
+parse-display = { version = "0.10.0", optional = true, default-features = false, features = [] }
```
DDtKey, line 57: `0.10.0` has changed MSRV
Let's use `0.9.1` for now, because `testcontainers` uses this too.

I'll update it once MSRV of `testcontainers` is bumped

src/openldap/mod.rs
DDtKey: I've left a comment here https://github.com/testcontainers/testcontainers-rs-modules-community/pull/180#discussion_r1723219292

]ToString] should be a good fit here

CommanderStorm: `IntoIterator` is much nicer to read. That was the idea that I was missing/not seeing 👍🏻

src/openldap/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+    pub fn with_ldap_password_hash(mut self, password_hash: PasswordHash) -> Self {
+        self.env_vars.insert(
+            "LDAP_PASSWORD_HASH".to_owned(),
+            match password_hash {
+                PasswordHash::SSHA => "{SSHA}".to_owned(),
+                PasswordHash::SHA => "{SHA}".to_owned(),
+                PasswordHash::SMD5 => "{SMD5}".to_owned(),
+                PasswordHash::MD5 => "{MD5}".to_owned(),
+                PasswordHash::CRYPT => "{CRYPT}".to_owned(),
+                PasswordHash::CLEARTEXT => "{CLEARTEXT}".to_owned(),
```
DDtKey, line 236: Minor, but we also can utilize parse display here:

```rs
#[derive(parse_display::Display)]
#[display("{{{}}}")] // it's escaped curly braces
enum PasswordHash { ... }
```

DDtKey (approved):

DDtKey:
Thank you for quick changes! ❤️

src/openldap/mod.rs
CommanderStorm: The reason why I did not include this is that I thought that this is too obvuscated.
But fair, is included ^^

Cargo.toml
CommanderStorm: have added a commit (includes a `TODO` comment why this is not up to date and when this can be updated ^^)

DDtKey (approved):

PR #182, opened 2024-08-19
feat: implement `with_init_sql` for `Postgres`, `Mysql` and `Mariadb`

### LINKED ISSUE #164: initSQL for postgres container

When running a test container, sometimes you might want to create a schema, user, role on the database before running your tests. 

I believe you we have the `.with_user`, and `.with_password` options. Is it possible to have a `.with_initsql` method on the Postgres implementation that executes arbitrary SQL scripts on the database on start up?

### Image reference

[dev-dependencies]
testcontainers = "0.20.0"
testcontainers-modules = { version = "0.8.0", features = ["postgres"] }

CommanderStorm:
Resolves #164

~The implementation is a bit hacky, am open to better approaches or other feedback ^^~

The implementation is pretty straigtforeward and similar to https://github.com/testcontainers/testcontainers-rs-modules-community/pull/217

### DIFF
--- a/src/mariadb/mod.rs
+++ b/src/mariadb/mod.rs
@@ -38,7 +38,7 @@ impl Mariadb {
     /// ```
     /// # use testcontainers_modules::mariadb::Mariadb;
     /// let mariadb_image = Mariadb::default().with_init_sql(
-    ///     "CREATE EXTENSION IF NOT EXISTS hstore;"
+    ///     "CREATE TABLE foo (bar varchar(255));"
     ///         .to_string()
     ///         .into_bytes(),
     /// );
--- a/src/mariadb/mod.rs
+++ b/src/mariadb/mod.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 
-use testcontainers::{core::WaitFor, CopyToContainer, Image};
+use testcontainers::{core::WaitFor, CopyDataSource, CopyToContainer, Image};
 
 const NAME: &str = "mariadb";
 const TAG: &str = "11.3";
@@ -27,7 +27,7 @@ const TAG: &str = "11.3";
 /// [`MariaDB docker image`]: https://hub.docker.com/_/mariadb
 #[derive(Debug, Default, Clone)]
 pub struct Mariadb {
-    init_sqls: Vec<CopyToContainer>,
+    copy_to_sources: Vec<CopyToContainer>,
 }
 
 impl Image for Mariadb {
@@ -55,17 +55,17 @@ impl Image for Mariadb {
         ]
     }
     fn copy_to_sources(&self) -> impl IntoIterator<Item = &CopyToContainer> {
-        &self.init_sqls
+        &self.copy_to_sources
     }
 }
 impl crate::InitSql for Mariadb {
-    fn with_init_sql(mut self, init_sql: impl ToString) -> Self {
-        let init_vec = init_sql.to_string().into_bytes();
+    fn with_init_sql(mut self, init_sql: impl Into<CopyDataSource>) -> Self {
         let target = format!(
             "/docker-entrypoint-initdb.d/init_{i}.sql",
-            i = self.init_sqls.len()
+            i = self.copy_to_sources.len()
         );
-        self.init_sqls.push(CopyToContainer::new(init_vec, target));
+        self.copy_to_sources
+            .push(CopyToContainer::new(init_sql.into(), target));
         self
     }
 }
@@ -83,7 +83,11 @@ mod tests {
     fn mariadb_with_init_sql() -> Result<(), Box<dyn std::error::Error + 'static>> {
         use crate::InitSql;
         let node = MariadbImage::default()
-            .with_init_sql("CREATE TABLE foo (bar varchar(255));")
+            .with_init_sql(
+                "CREATE TABLE foo (bar varchar(255));"
+                    .to_string()
+                    .into_bytes(),
+            )
             .start()?;
 
         let connection_string = &format!(
@@ -93,10 +97,10 @@ mod tests {
         );
         let mut conn = mysql::Conn::new(mysql::Opts::from_url(connection_string).unwrap()).unwrap();
 
-        let rows = conn.query("INSERT INTO foo(bar) VALUES ('blub')").unwrap();
+        let rows: Vec<String> = conn.query("INSERT INTO foo(bar) VALUES ('blub')").unwrap();
         assert_eq!(rows.len(), 0);
 
-        let rows = conn.query("SELECT bar FROM foo").unwrap();
+        let rows: Vec<String> = conn.query("SELECT bar FROM foo").unwrap();
         assert_eq!(rows.len(), 1);
         Ok(())
     }
--- a/src/mariadb/mod.rs
+++ b/src/mariadb/mod.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 
-use testcontainers::{core::WaitFor, Image};
+use testcontainers::{core::WaitFor, CopyToContainer, Image};
 
 const NAME: &str = "mariadb";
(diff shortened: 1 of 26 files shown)


src/postgres/mod.rs
```diff
@@ -61,6 +67,52 @@ impl Postgres {
...
+            .expect("we created files with filenames, not directories");
+        let container_path = format!("/docker-entrypoint-initdb.d/{file_name}",);
+        self.init_mounts.push(Mount::bind_mount(
+            init_path
+                .to_str()
+                .expect("the path we put is a valid string"),
+            container_path,
+        ));
+        Ok(self)
+    }
```
DDtKey, line 115: The similar approach can be implemented for some others DB modules (e.g mysq, mariadb)

Also, important part here: it uses blocking API and might be called from async code.

I think we need to have more general interface for these modules, with ability to run async and sync (we can spawn a blocking task at least)

I can suggest to extract a trait for RDBMS (e.g `InitSql` and `InitSqlAsync`). Also this common module may have different strategies: `TempFile`, `Sqlx` and etc. 
For now we can implement only the one on top of `TempFile`, it will create a temp directory/file and return a structure, like `TemporaryScript { tempdir: TempDir, path: PathBuf }`
So `Postgres` module will need only to mount this and keep the structure for it's lifetime


___
Probably it's overwhelming comment and not well described. Feel free to clarify anything and share your opinion

DDtKey:
Thank you for the contribution! ❤️ 
Left a couple of suggestions that could significantly expand the capabilities of the crate, but they need to be discussed and possibly postponed

src/postgres/mod.rs
```diff
@@ -70,7 +122,11 @@ impl Default for Postgres {
         env_vars.insert("POSTGRES_USER".to_owned(), "postgres".to_owned());
         env_vars.insert("POSTGRES_PASSWORD".to_owned(), "postgres".to_owned());
 
-        Self { env_vars }
+        Self {
+            env_vars,
+            init_mounts: Vec::new(),
+            tempdir: Arc::new(tempfile::tempdir().ok()),
```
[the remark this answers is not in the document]
CommanderStorm: Good point. Did not think about this, moving tempdir-cretaition into `with_init_sql` also improves the error handling in `with_init_sql`..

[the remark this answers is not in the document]
CommanderStorm: (lets split this into two parts and discuss if async is needed here instead of the other comment ^^)

https://docs.rs/tempfile/latest/tempfile/ does not offer an async API.
This may be because of two reasons:
- [nobody has requested it yet](https://github.com/Stebalien/tempfile/issues?q=is%3Aissue+async+)
- creating a directory is realy quick => having the async around this might be slower due to the overhead.

Also, since AsyncDrop is not a thing... https://github.com/sunsided/async-tempfile-rs/issues/1 any implemntation needs to be sync anyway (at least for the drop)

src/postgres/mod.rs
CommanderStorm: Might have misunderstood something ^^

This could work similarly, to the [wait's-stratgy patern](https://github.com/testcontainers/testcontainers-rs/tree/main/testcontainers/src/core/wait)..

```rust
#[derive(Default)]
enum InitSql {
  #[default]
  None,
  TemporaryDirectory(TemporaryDirectoryStrategy), // deliberatley not a TempFile, as this way multiple files are supported
  Sqlx(SqlxStrategy),
}

impl InitSql{
  pub fn mount_temporary_files(TemporaryDirectoryStrategy)
}

// InitSqlAsync is the same but with async in all relevant names
```

Not quite sure what a `InitSqlStrategy` trait would look like.
Any draft that I came up with shifts a lot of the work onto the RDMS..

Running sqlx commands after something is healthy and mounting files is not quite similar.
=> Not sure if marrying them at the hip as above is a good idea.. `TemporaryDirectory` needs a base path to retunrn the mounts while `Sqlx` needs a healthy connectionstring/PGConnection/...

src/postgres/mod.rs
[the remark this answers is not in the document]
DDtKey: I don't suggest using async api of tempfile. I'm aware it doesn't provide this. But it's not recommended to use blocking calls within async runtime. So it would be nice to cover with `spawn_blocking`, but not necessary on current stage - we could just declare that the api is blocking to warn users

>     creating a directory is realy quick => having the async around this might be slower due to the overhead.

It's actually not really true, it depends on many factors. Highly depends on disk usage and file-system. 

Moreover, we not only create a directory, but also write a file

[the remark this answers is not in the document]
CommanderStorm: good point

DDtKey:
Now it's possible with `with_copy_to`: 

```rs
         image.with_copy_to(
            "/docker-entrypoint-initdb.d/init.sql",
            include_bytes!("initdb.sql").to_vec(),
        )
```

Which I found simple enough. 
So we can actually provide a guidance and probably shortcut for it.

CommanderStorm:
Yea that definitely cleans things up a ton. Thanks 🙏 
Sorry for the lacking progress, I was on vacation for the last month and am still in a bit of a catch-up mode work wise.

=> This PR might take a few weeks before I can actually invest the time it deserves

src/postgres/mod.rs
```diff
@@ -62,6 +67,17 @@ impl Postgres {
         self
     }
 }
+impl crate::InitSql for Postgres {
```
CommanderStorm, line 70: this is essentially the same code for every intialisation.
I don't know if there is a way to move this to the trait.

src/mariadb/mod.rs
```diff
@@ -69,6 +79,31 @@ mod tests {
         testcontainers::{runners::SyncRunner, ImageExt},
     };
 
+    #[test]
+    fn mariadb_with_init_sql() -> Result<(), Box<dyn std::error::Error + 'static>> {
+        use crate::InitSql;
```
CommanderStorm, line 84: @DDtKey the choice of making this be a trait has the effect that this has to be imported.

That is a bit aqward imo => are you sure that you prefer it this way?

DDtKey: Usually it's fine to import traits in Rust. Many extensions work this way.
The idea of trait here was to unify this modules and avoid the possible differences, while have a documentation in one place.


However, we have only several modules with such functionality and I'd say: let's just keep separate methods for each one. I mean let's go without trait for now

src/postgres/mod.rs
DDtKey: Let's drop the trait for now. Just in-place methods are ok. We can change in the future with fallback to trait

src/mariadb/mod.rs
```diff
@@ -27,10 +27,37 @@ const TAG: &str = "11.3";
...
-    /// (remove if there is another variable)
-    /// Field is included to prevent this struct to be a unit struct.
-    /// This allows extending functionality (and thus further variables) without breaking changes
-    _priv: (),
+    copy_to_sources: Vec<CopyToContainer>,
+}
+
+impl Mariadb {
+    /// Registers sql to be executed automatically when the container starts.
+    ///
```
DDtKey, line 35: Can we mention that the method has append semantic? (in others too, please)

```suggestion
    /// Registers sql to be executed automatically when the container starts.
    /// You can call the method multiple times to add multiple scripts, they will be added and not overridden.
    ///
```

DDtKey (approved):

src/mariadb/mod.rs
CommanderStorm: I think the wording here can be simplified a bit, but otherwise entirely correct 😉
(I am okay with changing this to your version, if you like)

```suggestion
    /// Registers sql to be executed automatically when the container starts.
    /// You can call the method multiple times to add (not override) scripts.
    ///
```

DDtKey (approved):

Issue #2, opened 2023-06-19
Improve documentation of the crate


DDtKey:
We need to enrich the crate with documentation

Depends on #1

```[tasklist]
### Tasks
- [x] Contribution guide - @ubiratansoares is assigned
- [x] Improve crate-level documentation (docs.rs)
- [ ] Document usage of each existing module
- [x] Align readme with documentation
- [x] New issue template
- [x] Activate `#![deny(missing_docs)]` at crate level to avoid missing documentation
```

DDtKey:
Two remaining tasks are:
- Document usage of each existing module
We have a number of modules which has been moved from `testcontainers` repo, most of them don't have a documentation. Need to actualize and write a documentation like this: https://github.com/testcontainers/testcontainers-rs-modules-community/blob/6a4d10acbedd19f54a505afd09b66343744b4a64/src/rabbitmq/mod.rs#L6-L29

- Contribution guide

I would be glad for any help

level-one-learner:
Okay I'll have a go at adding documentation for  the use of each module and submit a PR.

ubiratansoares:
@DDtKey Is there anyone assigned to Contribution Guidelines?  If not, I'd love helping with that! 🙂

DDtKey:
No, feel free to take on this task!
Thank you!🙏

DDtKey:
To actualize the state:
We still need to cover undocumented modules. But there is no need to document all of them at once.

I'm happy to accept PRs with documentation for 1 module at a time.

CommanderStorm:
to update the issue:
Most items now (assuming the tow PRs get merged) have doc-strings, except the following, where I don't quite know how to document this.

https://github.com/testcontainers/testcontainers-rs-modules-community/blob/1886aa12b343ff02a55418e2d269d2d8d166c864/src/consul/mod.rs#L31-L33

CommanderStorm:
@ruslansennov since you contributed this:

What would be a good doc-string for this aspect? I don't know enough about said docker container to document this

CommanderStorm:
🎉

CommanderStorm:
Thank you people for doing this initiative in the first place.
Testcontainers make testing much nicer ^^

PR #183, opened 2024-08-20
docs: documented all ports which containers at least expose

CommanderStorm:
This is part of the larger documentation overhall initaltive https://github.com/testcontainers/testcontainers-rs-modules-community/issues/2

This is breaking in the sense that some containers ports were `u16` before..
Don't know how we should handle this..

### DIFF
--- a/src/kafka/mod.rs
+++ b/src/kafka/mod.rs
@@ -7,9 +7,16 @@ use testcontainers::{
 
 const NAME: &str = "confluentinc/cp-kafka";
 const TAG: &str = "6.1.1";
-
-pub const KAFKA_PORT: u16 = 9093;
-const ZOOKEEPER_PORT: u16 = 2181;
+/// Port that the [`Kafka`] part of the container has internally
+/// Can be rebound externally via [`testcontainers::core::ImageExt::with_mapped_port`]
+///
+/// [`Kafka`]: https://kafka.apache.org/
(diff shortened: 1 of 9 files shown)


src/kafka/mod.rs
```diff
@@ -7,9 +7,16 @@ use testcontainers::{
...
 const NAME: &str = "confluentinc/cp-kafka";
 const TAG: &str = "6.1.1";
-
-pub const KAFKA_PORT: u16 = 9093;
-const ZOOKEEPER_PORT: u16 = 2181;
+/// Port that the [`Kafka`] part of the container has internally
+/// Can be rebound externally via [`testcontainers::core::ImageExt::with_mapped_port`]
+///
+/// [`Kafka`]: https://kafka.apache.org/
+pub const KAFKA_PORT: ContainerPort = ContainerPort::Tcp(9093);
```
DDtKey, line 14: Looks like `KAFKA_PORT` is only breaking change here. Others were private or not-documented. 

I'd suggest to extract this breaking change into separate PR, thus we will have 2 PRs:
- `docs: ...`  (compatible )
- `refactor!: use `ContainerPort` for exposed kafka port

DDtKey (commented):
Thank you for the contribution! I appreciate your help with the project 🚀

DDtKey (approved):

PR #184, opened 2024-08-20
docs: added docs for Apache ZooKeeper

CommanderStorm:
This is part of the work to get https://github.com/testcontainers/testcontainers-rs-modules-community/issues/2 resolved

Btw:
Hope the way I write PRs (a gaggle of unrelated PRs surrounding a PR) is not too stressfull on your side.

PR #186, opened 2024-08-21
fix: doc-test failures

CommanderStorm:
- I am soo sorry.
  Idk why, but apparenty I forgot to run `cargo test` on https://github.com/testcontainers/testcontainers-rs-modules-community/pull/184 while documenting this on my commute 👀 
- I also included the mongo-example patch


DDtKey (approved):

DDtKey:
Thanks!
I've updated the title, because it's `fix:` rather than `ci:` because we don't change the CI itself

PR #187, opened 2024-08-21
docs: enable `rustfmt` for doc comments

CommanderStorm:
`format_code_in_doc_comments` is a nightly `rustfmt` feature. (tracking issue: https://github.com/rust-lang/rustfmt/issues/3348)
I think this might be helpfull.

Usually, I would not consider this being an option, but given that already some other `rustfmt` feature are enabled... 😉

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -3,3 +3,4 @@ reorder_imports = true
 merge_derives = true
 imports_granularity = "Crate"
 group_imports = "StdExternalCrate"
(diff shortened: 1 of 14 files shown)


DDtKey (approved):
Usage of nightly features for rustfmt is quite common. Looks reasonable, let's enable

CommanderStorm:
I have split out the CI part to https://github.com/testcontainers/testcontainers-rs-modules-community/pull/188 to get rid of the warnings

PR #189, opened 2024-08-21
docs: fixed rustfmt violation in the `openldap` module

CommanderStorm:
So apparently, yesterday another fuckup made its way into master. Sorry about that.


DDtKey (approved):
It seems it was a coincidence that we added a new Rustfmt rule and didn't make sure it worked for openldap PR. 
Thanks for the fixing it

Issue #200, opened 2024-08-30
has someone gotten `k3s` to work?


bbortt:
### Feature request description

I am trying to deploy a pod into `k3s`. I've followed the [contained test](https://github.com/testcontainers/testcontainers-rs-modules-community/blob/a7fd7d0408495c1b3364e3c01a50b7a22ac1a79c/src/k3s/mod.rs#L170) (roughly), but fail to get it working. I've also considered the [`pod_api.rs` example](https://github.com/kube-rs/kube/blob/main/examples/pod_api.rs) over at `kube-rs/kube`.

**has someone gotten this to work?** I always end up with: `Service(hyper_util::client::legacy::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })))`.

this happens to be thrown in the `pods.create` line:

```rust
pub(crate) async fn deploy_argocd(client: Client) {
    let pods: Api<Pod> = Api::default_namespaced(client.clone());

    let argocd_pod: Pod =
        parse_yaml("tests/resources/k3s/argocd.pod.yml").expect("Failed to read ArgoCD deployment");

    pods.create(&PostParams::default(), &argocd_pod)
        .await
        .expect("Failed to deploy ArgoCD to K3s");
}
```

### Image reference

[`k3s` feature](https://github.com/testcontainers/testcontainers-rs-modules-community/blob/a7fd7d0408495c1b3364e3c01a50b7a22ac1a79c/src/k3s/mod.rs)

DDtKey:
Hi @bbortt 👋 
Could you say what version of `kube` do you use in your tests?

bbortt:
@DDtKey thanks for the quick response. I saw that the test suite works and I can reproduce that. but it does only read pods, not create any new.. as soon as I go into this direction I end up with the mentioned error..

```toml
[dev-dependencies]
assert_cmd = "2.0.16"
k8s-openapi = { version = "0.22.0", features = ["latest"] }
kube = { version = "0.93.1", features = ["derive"] }
kube-derive = "0.93.1"
predicates = "3.1.2"
rustls = "0.23.12"
schemars = "0.8.21"
testcontainers-modules = { version = "0.9.0", features = [ "hashicorp_vault", "postgres"] }
```

DDtKey:
I guess you've figured out the reason and described in https://github.com/testcontainers/testcontainers-rs-modules-community/pull/202

[...]

PR #202, opened 2024-08-31
docs(#200): add k3s pod api example

bbortt:
reproducer for #200.. let's see if I am the honk 😜


bbortt:
whoever approved the workflow beforehand; thanks. can you please do it again?

I was unable to build the project locally up until now, but managed to do so. for anyone running into problems with `rdkafka` on windows ("this is not an x32 architecture"), try adding the `cmake` build:

```toml
rdkafka = { version = "0.36", features = ["cmake-build"] }
```

src/k3s/mod.rs
```diff
@@ -224,6 +222,84 @@ mod tests {
...
+            "apiVersion": "v1",
+            "kind": "Pod",
+            "metadata": { "name": "blog" },
+            "spec": {
+                "containers": [{
+                  "name": "blog",
+                  "image": "clux/blog:0.1.0"
+                }],
+            }
+        }))?;
```
DDtKey, line 252: I'd suggest to move this into [/examples](https://github.com/testcontainers/testcontainers-rs-modules-community/tree/main/examples) instead of tests. Because it uses some custom image reference and it would be extra waiting time and load for CI

DDtKey (commented):
Thank you for the contribution and example! 🙏 

>whoever approved the workflow beforehand; thanks. can you please do it again?

Done ✅

src/k3s/mod.rs
bbortt: are examples execute in ci as well? in a later stage, somewhere?

bbortt: I am asking, because a sample that does not run is not worth it imo

DDtKey: No, examples are compiled but not executed. They mostly needed as a demonstration - but yes, that's valid point.

I just don't want CI to depend on some random image (e.g `clux/blog:0.1.0`). We can add the test, but with some more predictable image

bbortt: > I just don't want CI to depend on some random image

understood.

how about re-using something like the postgres image - that's being used anyway? I'll have a look at it! or maybe some very small minimal image.

bbortt:
@DDtKey here come some more findings..

I did only manage to run multiple tests using `k3s` in _serial_, not parallel mode. I therefore had to add the `test_serial` create to the dev-dependencies.

if you don't do that and run the tests in `parallel`, you'll run into the exact same [error that happened in the last "test (stable)" job](https://github.com/testcontainers/testcontainers-rs-modules-community/actions/runs/10683032900/job/29611669793?pr=202#step:8:629). it's quit reproducible! I think in around 10 times I ran it locally it never passed.

on the other hand, the error initially reported in #200 did no longer occur. that's good 😉 however, that also means I have to double-check the code in my other project - that's less good haha

finally, I've used the [`busybox` image](https://hub.docker.com/_/busybox/) for the test. I think it's less "random" and quit known. it's also only around 1.5 MB in size, so sufficiently small to not really slow down the ci imo. what do you think of it?

and finally, thanks for all your patience! it's my first rust project contribution ❤️ I've already learned a lot again.

PR #207, opened 2024-09-15
feat: support databend module

### LINKED ISSUE #212: [feat] Add support for Databend

[Databend](https://github.com/datafuselabs/databend), built in Rust, is an open-source cloud data warehouse, we can support it in testcontainers-rs.

### Image reference

_No response_

hantmac:
Support [databend](https://github.com/datafuselabs/databend) module.

Closes #212


src/databend/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+mod tests {
+    use databend_driver::{Client, Connection};
+    use testcontainers::ImageExt;
+    use crate::{databend::Databend as DatabendImage, testcontainers::runners::AsyncRunner};
+
+    #[tokio::test]
+    async fn test_databend() {
+        let databend = DatabendImage::default()
+            .with_env_var("QUERY_DEFAULT_USER","databend")
+            .with_env_var("QUERY_DEFAULT_PASSWORD","databend")
```
DDtKey, line 75: I think we should provide some defaults and builder-like methods to override these envs (e.g `with_default_user(..)`)

WDYT?

Cargo.toml
```diff
@@ -22,6 +22,7 @@ clickhouse = ["http_wait"]
 cncf_distribution = []
 consul = []
 dynamodb = []
+databend = []
```
DDtKey, line 25: It's required to enable http strategy for this module explicitly

```suggestion
databend = ["http_wait"]
```

DDtKey (commented):
Thank you for the contribution! 🙏 

I've left a few comments, please take a look

src/databend/mod.rs
hantmac: Because the `core/images.rs` does not provide the  `with_query_default_user` method, so we can't impl it in `impl Image for Databend`. Do you have any suggesstions?

DDtKey: I meant to `impl Databend { pub fn … }`
It shouldn’t be a part of common trait.

You can check some other modules, for example: https://github.com/testcontainers/testcontainers-rs-modules-community/blob/f2bf770291d4806c1eb6aad1d0a89874b1494995/src/postgres/mod.rs#L36

DDtKey: Also, you can impl Default with these env vars. And mention the default user and password in the doc comment for Databend. But user will be able to override it

hantmac: @DDtKey Thanks! I try to add the code.Plz review.

Cargo.toml
DDtKey: This one is still relevant. Also please apply rustfmt and LGTM!

Thank you!

DDtKey (approved):

DDtKey:
Thank you!🙏

src/databend/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    fn expose_ports(&self) -> &[ContainerPort] {
+        &[DATABEND_PORT]
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use databend_driver::{Client, Connection};
```
DDtKey, line 92:
```suggestion
    use databend_driver::Client;
```

src/databend/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    use testcontainers::ImageExt;
+
+    use crate::{databend::Databend as DatabendImage, testcontainers::runners::AsyncRunner};
+
+    #[tokio::test]
+    async fn test_databend() {
+        let databend = DatabendImage::default().start().unwrap();
```
DDtKey, line 99:
```suggestion
        let databend = DatabendImage::default().start().await.unwrap();
```

src/databend/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+        let http_port = databend.get_host_port_ipv4(8000).unwrap();
```
DDtKey, line 100:
```suggestion
        let http_port = databend.get_host_port_ipv4(8000).await.unwrap();
```

DDtKey:
`?` won't help, because the function doesn't return `Result`. The issue is that it's async function and await wasn't called. See the suggestions above

src/databend/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    fn expose_ports(&self) -> &[ContainerPort] {
+        &[DATABEND_PORT]
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use databend_driver::{Client};
```
DDtKey, line 92:
```suggestion
    use databend_driver::Client;
```

src/databend/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    use crate::{databend::Databend as DatabendImage, testcontainers::runners::AsyncRunner};
+
+    #[tokio::test]
+    async fn test_databend() {
+        let databend = DatabendImage::default().start().await.unwrap();;
```
DDtKey, line 98:
```suggestion
        let databend = DatabendImage::default().start().await.unwrap();
```

DDtKey (approved):

PR #222, opened 2024-09-25
docs: disallow missing docs on crate level and allow where necessary

CommanderStorm:
This PR is more for discussion which of these two approches you would like to prefer:

- deny missing docs now, work on the allows for a while
- work on the allows for a while, deny missing docs later

the advantage of denying now is that regressions (new modules being added without docs) is not possible anymore.

For example, https://github.com/testcontainers/testcontainers-rs-modules-community/pull/222/commits/98429e430398fe43cee720ee158150b9af89c7d8 would have been caught in https://github.com/testcontainers/testcontainers-rs-modules-community/pull/218 ^^


CommanderStorm:
Okay, lint-reasons were only [stabilised in 1.81](https://github.com/rust-lang/rust/pull/120924) (not 1.75)
=> they cannot be used.

Will switch to comments

DDtKey (approved):
Thank you! ❤️
I find this reasonable - we can cover step by step removing allows. But we must enforce it for new changes

PR #233, opened 2024-10-14
feat: add `Pulsar` module

alesharik:
This PR adds module for [Apache Pulsar](https://pulsar.apache.org/)

### DIFF
--- a/src/pulsar/mod.rs
+++ b/src/pulsar/mod.rs
@@ -37,8 +37,7 @@ pub struct Pulsar {
 
 impl Default for Pulsar {
     /**
-     * Starts an in-memory instance in dev mode, with horrible token values.
-     * Obviously not to be emulated in production.
+     * Creates new standalone pulsar container, with `/pulsar/data` as temporary volume
      */
     fn default() -> Self {
         Self {
--- a/src/pulsar/mod.rs
+++ b/src/pulsar/mod.rs
@@ -37,7 +37,7 @@ pub struct Pulsar {
 
 impl Default for Pulsar {
     /**
-     * Creates new standalone pulsar container, with `/pulsar/data` as temporary volume
+     * Creates new standalone pulsar container, with `/pulsar/data` as a temporary volume
      */
     fn default() -> Self {
         Self {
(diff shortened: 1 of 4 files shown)


src/pulsar/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+impl Pulsar {
+    /// Add configuration parameter to Pulsar `conf/standalone.conf`
+    pub fn with_config(mut self, name: impl Into<String>, value: impl Into<String>) -> Self {
+        self.env
+            .insert(format!("PULSAR_PREFIX_{}", name.into()), value.into());
+        self
+    }
```
DDtKey, line 58: The documentation might be a bit confusing, because it's implemented utilizing ENV variables, while use can utilize mounting or `CopyTo` functionality for configuration file

Perhaps it should be `with_config_env` method? mentioning that it's an env variable with predefined prefix in accordance with https://pulsar.apache.org/docs/next/getting-started-docker/

WDYT?

src/pulsar/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+    data_mount: Mount,
+    env: BTreeMap<String, String>,
+    admin_commands: Vec<Vec<String>>,
+}
+
+impl Default for Pulsar {
+    /**
+     * Starts an in-memory instance in dev mode, with horrible token values.
+     * Obviously not to be emulated in production.
+     */
```
DDtKey, line 42:
```suggestion
```

a nit: I think it's excessive,  `testcontainers` are in general not indented to be used in production and default insecure values are expected for most of the containers.

And additionally: it's instantiation, without startup. User can override all tokens and etc before starting the container

DDtKey:

It looks good to me, just couple of suggestions

src/pulsar/mod.rs
alesharik: Good point! And also that clarifies that configuration is done with py script, which sometimes behaves differently that writing config directly into file

DDtKey (approved):

PR #255, opened 2024-12-18
feat: add "pro" variant for LocalStack image

the-wondersmith:
PR adds a "pro" variant of the existing [LocalStack](https://docs.rs/testcontainers-modules/latest/testcontainers_modules/localstack/struct.LocalStack.html) image.

### DIFF
--- a/src/localstack/mod.rs
+++ b/src/localstack/mod.rs
@@ -1,4 +1,8 @@
+pub use pro::LocalStackPro;
 use testcontainers::{core::WaitFor, Image};
+
+/// LocalStack Pro
+pub mod pro;
 
 const NAME: &str = "localstack/localstack";
 const TAG: &str = "3.8";
--- a/src/localstack/mod.rs
+++ b/src/localstack/mod.rs
@@ -5,12 +5,12 @@ use testcontainers::{core::WaitFor, Image};
 pub mod pro;
 
 const NAME: &str = "localstack/localstack";
-const TAG: &str = "3.8";
+const TAG: &str = "3.0";
 const DEFAULT_WAIT: u64 = 3000;
 
 /// This module provides [LocalStack](https://www.localstack.cloud/) (Community Edition).
 ///
-/// Currently pinned to [version `3.8`](https://hub.docker.com/layers/localstack/localstack/3.8/images/sha256-fc9a03f14f4668f5d874eadb77e5da5461ce735e2ce86b77b0056606c0677dba?context=explore)
+/// Currently pinned to [version `3.0`](https://hub.docker.com/layers/localstack/localstack/3.0/images/sha256-73698e485240939490134aadd7e429ac87ff068cd5ad09f5de8ccb76727c13e1?context=explore)
 ///
 /// # Configuration
 ///
--- a/src/localstack/mod.rs
+++ b/src/localstack/mod.rs
@@ -1,12 +1,12 @@
 use testcontainers::{core::WaitFor, Image};
 
 const NAME: &str = "localstack/localstack";
-const TAG: &str = "3.0";
+const TAG: &str = "3.8";
 const DEFAULT_WAIT: u64 = 3000;
 
 /// This module provides [LocalStack](https://www.localstack.cloud/) (Community Edition).
 ///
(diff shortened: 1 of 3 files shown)


src/localstack/mod.rs
```diff
@@ -1,12 +1,16 @@
+pub use pro::LocalStackPro;
 use testcontainers::{core::WaitFor, Image};
 
+/// LocalStack Pro
+pub mod pro;
+
 const NAME: &str = "localstack/localstack";
-const TAG: &str = "3.0";
+const TAG: &str = "3.8";
```
DDtKey, line 8: We usually consider version bumps as a breaking change, because unexpected regressions may happen even for minor updates

If it's possible to just update version without changes from our side - user always can do that by specifying particular version and that's even recommended approach

src/localstack/mod.rs
DDtKey, line 1: I wonder if we really need a separate image impl, or it can be dynamic solution 🤔

E.g:
`Localstack::new(..).pro()`

However it's a separate solution and two separate definitions align well with other modules (like `redis`)
So I don't mind, but also don't have strict opinion in this case

DDtKey (commented):
Looks good to me, a few comments to discuss only

Thank you! 🙏

src/localstack/mod.rs
the-wondersmith: I *did* consider something like a `.pro()` method, (among other options), but there *are* a few assumptions/implications that are slightly different between the "regular" and "pro" image variants - for example the auth token. Arguably, the `LocalStackPro` variant should *require* an `auth_token` value, as otherwise you'd expect a user to just use the `LocalStack` variant. *However*, the pro images include a method to disable the activation of the pro features (`ACTIVATE_PRO=0`), so... here's where we ended up 😅.

It *is* technically a separate solution, so a separate definition seems to make sense.

src/localstack/mod.rs
the-wondersmith: Just to get it on the record: there's an argument to be made that the pinned image versions should be kept *reasonably* up to date, (maybe on some kind of standard cadence that'd reflected in the crate version? 🤔), and I'd absolutely be one to make that argument.

You're right though, about it being a breaking change. Since this PR is intended to be a patch version bump at most, I just went ahead and reverted the version bump and updated the PR description 🙂.

src/localstack/mod.rs
```diff
@@ -1,12 +1,16 @@
...
+pub mod pro;
+
 const NAME: &str = "localstack/localstack";
 const TAG: &str = "3.0";
 const DEFAULT_WAIT: u64 = 3000;
 
 /// This module provides [LocalStack](https://www.localstack.cloud/) (Community Edition).
 ///
-/// Currently pinned to [version `3.0`](https://hub.docker.com/layers/localstack/localstack/3.0/images/sha256-73698e485240939490134aadd7e429ac87ff068cd5ad09f5de8ccb76727c13e1?context=explore)
+/// Currently pinned to [version `3.8`](https://hub.docker.com/layers/localstack/localstack/3.8/images/sha256-fc9a03f14f4668f5d874eadb77e5da5461ce735e2ce86b77b0056606c0677dba?context=explore)
```
DDtKey, line 13: we reverted version update, so documentation shouldn't be changed too here

DDtKey (approved):
Looks good to me

src/localstack/mod.rs
the-wondersmith: Good catch 😅. Fixed.

DDtKey:
Yes, I guess MSRV was bumped as part of path/minor version of the dependency.

We can either stick it to particular version, or consider MSRV update, but as a breaking change.

Using specific version of dependency may be a pain for our users, if they already updated it
I think we can consider update of MSRV in separate PR

the-wondersmith:
> Yes, I guess MSRV was bumped as part of path/minor version of the dependency.
> 
> We can either stick it to particular version, or consider MSRV update, but as a breaking change.
> 
> Using specific version of dependency may be a pain for our users, if they already updated it I think we can consider update of MSRV in separate PR

What's that mean for this PR? Does it need to wait until CI can be fixed?

PR #265, opened 2025-01-07
ci: utilize cargo-hack partitions to speed up tests

DDtKey:
We have many modules under separate features and CI takes too long. 
In order to improve it, we can try to utilize partitions and run several machines at once.

Also, it may solve issue with disk space required.


DDtKey:
I find this useful:
- stable tests take 12min max
- nightly takes just 3-5 minutes


Used to be ~25-30 minutes before.
So it's significantly better IMO (x2 at least)

PR #268, opened 2025-01-12
feat: allow enabling jetstream for nats

cedricziel:
NATS has a built-in persistence engine called JetStream which enables messages to be stored and replayed at a later time.

https://docs.nats.io/nats-concepts/jetstream

### DIFF
--- a/src/nats/mod.rs
+++ b/src/nats/mod.rs
@@ -38,8 +38,17 @@ impl NatsServerCmd {
         self
     }
 
-    // not having docs here is currently allowed to address the missing docs problem one place at a time. Helping us by documenting just one of these places helps other devs tremendously
-    #[allow(missing_docs)]
+    /// Enable JetStream in the Nats server to use the built-in persistence
+    /// features of NATS.
+    ///
+    /// See: https://docs.nats.io/nats-concepts/jetstream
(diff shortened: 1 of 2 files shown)


src/nats/mod.rs
```diff
@@ -35,6 +37,13 @@ impl NatsServerCmd {
         self.pass = Some(password.to_owned());
         self
     }
+
+    // not having docs here is currently allowed to address the missing docs problem one place at a time. Helping us by documenting just one of these places helps other devs tremendously
+    #[allow(missing_docs)]
```
DDtKey, line 42: Please, just cover the method with documentation

We allow `missing_docs` only for existing (old, not covered) code. For all new code we add the documentation, since the target is to cover all parts

DDtKey (commented):

Looks good to me, but just one comment

cedricziel:
@DDtKey I added a small blurb about the JetStream feature in NATS - let me know if that works :)

DDtKey (approved):

cedricziel:
I think the test fails because a different module, not because of the changes in this PR

DDtKey:
Yes, `gitea` module seems to be flaky. 

PR #272, opened 2025-01-13
feat: add Foundry Anvil image impl

suchapalaver:
# Community Testcontainers Implementation for [Foundry Anvil](https://book.getfoundry.sh/anvil/)

This is a community implementation of the [Testcontainers](https://testcontainers.org/) interface for [Foundry](https://github.com/foundry-rs/foundry) [Anvil](https://book.getfoundry.sh/anvil/).

It is not officially supported by Foundry, but it is a community effort to provide a more user-friendly interface for running Anvil inside a Docker container.

The endpoint of the container is intended to be injected into your provider configuration, so that you can easily run tests against a local Anvil instance. See the `test_anvil_node_container` test in `src/anvil/mod.rs` for an example.


suchapalaver:
The CI failures do not appear to be caused by the changes in this PR.

DDtKey:
@suchapalaver Yes, `gitea` and `oracle` tests become flaky in CI, not related to this PR.

src/anvil/mod.rs
```diff
@@ -0,0 +1,139 @@
+use std::borrow::Cow;
+
+use testcontainers::{
+    core::{ContainerPort, WaitFor},
+    Image,
+};
+
+const NAME: &str = "ghcr.io/foundry-rs/foundry";
+const TAG: &str = "latest";
```
DDtKey, line 9: We prefer to avoid  `latest` tag as it can easily lead to a module being broken (e.g. an incompatible version is released).

Users always can override it with [ImageExt::with_tag](https://docs.rs/testcontainers/0.23.1/testcontainers/core/trait.ImageExt.html#tymethod.with_tag) in their code.

DDtKey (commented):

Just one comment and LGTM!

src/anvil/mod.rs
suchapalaver: Good point! The [Foundry package tags](https://github.com/foundry-rs/foundry/releases) are a bit unfriendly in this regard so I've gone with `stable` but with the sha as well. Hope that does the job.

suchapalaver:
@DDtKey thanks for your help on this one!

Issue #273, opened 2025-01-15
Fix gitea and oracle CI


suchapalaver:
### Describe the bug

At time of writing CI is failing on the `gitea` and `oracle` features.

### To Reproduce

Contribute a PR to this repo. 

You can also observe the behavior by visiting the Actions on a recently opened PR.

### Expected behavior

CI checks on code in `main` should pass.

DDtKey:
`gitea` and `oracel` modules have flaky tests in CI, they may pass or fail periodically.

It seems the situation has gotten worse and I would appreciate any contribution that will help to eliminate flakiness.

Thank tou @suchapalaver for filling the issue!

PR #276, opened 2025-01-23
perf(postgres)!: set `fsync=off` by default

luca-iachini:
This PR updates the Postgres container to set `fsync=off`.
Disabling `fsync` enhances test performance by minimizing disk I/O, as data durability guarantees are typically unnecessary during tests.


src/postgres/mod.rs
```diff
@@ -127,9 +127,14 @@ impl Image for Postgres {
...
         &self.env_vars
     }
+
     fn copy_to_sources(&self) -> impl IntoIterator<Item = &CopyToContainer> {
         &self.copy_to_sources
     }
+
+    fn cmd(&self) -> impl IntoIterator<Item = impl Into<std::borrow::Cow<'_, str>>> {
+        vec!["-c", "fsync=off"]
+    }
```
DDtKey, line 137: Let’s make it configurable instead, as we might change behavior that someone relies on.

We could add a method like  `with_fsync(bool)`  instead

DDtKey (commented):
Thanks for the contribution 🙏 

I don't mind at all, just one suggestion to preserve compatibility and to make it explicit

src/postgres/mod.rs
luca-iachini: Thanks for your feedback! I believe that `fsync=off` should be the default setting. Disabling `fsync` doesn't prevent data from being written to disk. It reduces I/O overhead, with the only potential downside of data loss in the event of a system crash (which is not a concern for tests because they are ephimeral). Since data is still written to disk, the volume remains accessible for inspection if needed. In my experience, It reduces the execution time of tests significantly. 
By the way, Java Testcontainers already use this configuration by default, as shown in this [issue](https://github.com/testcontainers/testcontainers-java/issues/1256). I believe this library should provide meaningful defaults optimized for testing.
Let me know if that makes sense to you.

DDtKey: > Disabling fsync doesn't prevent data from being written to disk.

Sure, I'm aware of it, but my example above is still valid - 

> e.g someone uses volumes and persists state for debugging

Sometimes, CI is flaky and you want to check the dump of DB after test-run. You can utilize volumes and GH artifacts for example, but if `fsync=off` is enabled - you can't be sure DB isn't corrupted because it could be due to system crash.

Another use-case of testcontainers I know is chaos testing: with restarts, container crashes and network delays and etc (to ensure their application can survive these scenarios and service is self-healing).  
We can't imagine all the variants.

__ 

**But let me to clarify** - **I don't mind of making it the default**. It makes perfect sense to me because cases I described above are specific while usually tests don't need this enabled. However **it must be considered breaking change** IMO.  

Therefore, changing default behavior requires new major release (0.12+ I mean)

And also it still would be nice to be able to control the behavior with simple interface. 


Thus, we can:
- introduce method right now and release within patch version
- prepare PR to change the default and release the change with 0.12

Or
- add method, keep `fsync=off` default, but release might be postponed

DDtKey:
Btw, don't worry about failed `gitea` tests - they're [flaky right now](https://github.com/testcontainers/testcontainers-rs-modules-community/issues/273). It's known issue, but not a blocker for your PR

src/postgres/mod.rs
DDtKey: > That's my point — keeping fsync=off as the default covers almost 99% of cases.

And as I said - I agree. But 99% doesn't mean the change is compatible for 100% use-cases - we can't know.

My opinion is still the same - **let's make it a `fsync=off` a default, but provide ability to control this flag**, it's deadly simple to do - but can be useful if someone relies on it. 

I don't understand the issue of marking this as a breaking change, because it's health and proper version evolution of module. Otherwise there is a risk of breaking someones behavior

___
> For this to be a real concern, you'd have to experience OS or hardware failures on the runner itself.

That's not really true,  when postgres containerized - OS is part of image, there is no need to encounter runner's crash. 

> I’m not too familiar with chaos testing, so I could be wrong here. (I found an example where they use the Java Testcontainers wich has fsync=off: [Supercharge Your Spring Boot Testcontainers and Chaos Engineering](https://dev.to/aaravjoshi/supercharge-your-spring-boot-testcontainers-and-chaos-engineering-for-bulletproof-apps-1l3l)).

It depends on casy-by-case. If your chaos testing involves tools like [pumba](https://github.com/alexei-led/pumba) and you kills containers randomly - it might be important to keep `fsync` enabled to ensure failovers and etc works properly

luca-iachini: I've pushed the change to allow configuring it, as you suggested—it was indeed straightforward.

Regarding the other points, I see things differently. Since Docker uses the host kernel and postgres runs as an isolated process on the host. 

Anyway, I think it should be okay now.

DDtKey: >I've pushed the change to allow configuring it, as you suggested—it was indeed straightforward.

> Regarding the other points, I see things differently. Since a Docker container uses the host kernel and postgres runs as an isolated process on the host.

That doesn't mean that OS can't crush within container, not the host system. Docker image still contains

DDtKey (approved):

src/postgres/mod.rs
```diff
@@ -91,6 +92,12 @@ impl Postgres {
             .push(CopyToContainer::new(init_sql.into(), target));
         self
     }
+
+    /// Enables fsync for the Postgres instance.
```
CommanderStorm, line 96: We might consider copying the postgres docs on this feature ^^
(feel free to shorten/ignore this input)

```suggestion
    /// Enables `fsync` for the Postgres instance.
    /// 
    /// If this option is on, the PostgreSQL server will try to make sure that updates are physically written to disk, by issuing `fsync` system calls or various equivalent methods (see [wal_sync_method](https://www.postgresql.org/docs/8.1/runtime-config-wal.html#GUC-WAL-SYNC-METHOD)).
    /// This ensures that the database cluster can recover to a consistent state after an operating system or hardware crash.
    /// 
    /// However, using `fsync` results in a performance penalty:
    /// When a transaction is committed, PostgreSQL must wait for the operating system to flush the write-ahead log to disk.
    /// When `fsync` is disabled, the operating system is allowed to do its best in buffering, ordering, and delaying writes.
    /// This can result in significantly improved performance.
    /// However, if the system crashes, the results of the last few committed transactions may be lost in part or whole.
    /// In the worst case, unrecoverable data corruption may occur.
    /// Crashes of the database software itself are not a risk factor here.
    /// Only an operating-system-level crash creates a risk of corruption.
    /// 
    /// Due to the risks involved, there is no universally correct setting for `fsync`.
    /// Some administrators always disable `fsync`, while others only turn it off during initial bulk data loads, where there is a clear restart point if something goes wrong.
    /// Others always leave `fsync` enabled.
    /// The default is to enable `fsync`, for maximum reliability.
    /// If you trust your operating system, your hardware, and your utility company (or your battery backup), you can consider disabling `fsync`.
```

DDtKey: I think it's excessive, we can provide link to the documentation instead (and users are responsible to ensure they are reading postgres documentation in accordance with their version)

CommanderStorm:
```suggestion
    /// Enables fsync for the Postgres instance.
    /// 
    /// See <https://www.postgresql.org/docs/8.1/runtime-config-wal.html> for further context
```

or alternative

```suggestion
    /// Enables [the fsync-setting](https://www.postgresql.org/docs/8.1/runtime-config-wal.html) for the Postgres instance.
```

luca-iachini: @CommanderStorm Is there a specific reason why the link points to version 8.1 of the documentation?

DDtKey: I believe it should be either [11](https://www.postgresql.org/docs/11/runtime-config-wal.html) (because it's default version of image) or just [current](`https://www.postgresql.org/docs/current/runtime-config-wal.html`)

(no strong preference, both are fine)

CommanderStorm: (no particular reason for the version choice, that was chosen by Google)

PR #290, opened 2025-02-10
fix: dex config serialization

yuri-becker:
When implementing more complex test cases, i noticed i messed up a bit with config serialization. I hope it's okay to open another PR so shortly


DDtKey (approved):
That's ok, moreover we haven't released it yet 

Thank you!🙏

PR #303, opened 2025-03-01
feat(minio)!: support newest MinIO versions

quodlibetor:
It seems like at some point in the last 3 years MinIO switched from writing startup messages on stdou to stderr. I couldn't find a `WaitFor::either()` to keep the old version as the default, so I bumped it to the latest released version.

This update is necessary for folks to be able to use a modern or different tag with `ImageExt`.

### DIFF
--- a/src/minio/mod.rs
+++ b/src/minio/mod.rs
@@ -3,7 +3,7 @@ use std::{borrow::Cow, collections::HashMap};
 use testcontainers::{core::WaitFor, Image};
 
 const NAME: &str = "minio/minio";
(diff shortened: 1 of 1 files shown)


quodlibetor:
Updated the PR title to conventional commit, I think that's what failed all of CI.

quodlibetor:
I spoke too soon, this is an additional problem: https://github.com/testcontainers/testcontainers-rs-modules-community/pull/303

DDtKey (approved):

DDtKey:
Thank you! 🙏

PR #304, opened 2025-03-02
chore(deps): pin `chrono` as a dev-dep to fix `arrow-arith` compilation failure

quodlibetor:
arrow-arith bug report: https://github.com/apache/arrow-rs/issues/7196

You don't want to actually restrict the Chrono version since your deps need to choose their own workaround, but this will hopefully fix CI.

See #305 for an alternative.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,11 @@ testcontainers = { version = "0.23.2" }
 
 
 [dev-dependencies]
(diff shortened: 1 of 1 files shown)


DDtKey (approved):

DDtKey:
Accepting as separate PR due to well-written comment and explicit git history record, thank you! 🙏 

Deps will be updated with #306

PR #305, opened 2025-03-02
fix: Update databend-driver to widen allowed arrow versions

quodlibetor:
0.22.2 caused ancestor deps to not allow updating arrow, which in turn meant that we couldn't get the hotfix for the arrow-arith issues mentioned in #304.

This is an alternative, plausibly better fix. But I don't know what databend is so I'm not sure if there's a reason you were on an older version.


Cargo.toml
```diff
@@ -87,7 +87,7 @@ aws-sdk-s3 = "1.2.0"
 aws-sdk-sqs = "1.2.0"
 aws-types = "1.0.1"
 bollard = "0.18.1"
-databend-driver = "0.22.2"
+databend-driver = "0.26.1"
```
DDtKey, line 90: That's not enough, underlying issue is pinnded `chrono 0.4.39`

DDtKey (commented):
Thank you for the attempt to fix the issue 🙏 
I really appreciate your time

However there is a prepared update for all dev-deps, it's gonna be fixed within that batch of changes. 

Also I'm accepting your #304

PR #307, opened 2025-03-02
feat(surrealdb)!: support surrealdb v2

DDtKey:
This is incompatible update because requires appropriate client.

For the users of surreal v1 it's still possible to use an old version of client and tag customization


DDtKey:
Fails in CI not related to this change and will be fixed separately

PR #326, opened 2025-04-08
fix: lint error

tomers:
Fixes the following error:

error: use of deprecated associated function `aws_config::BehaviorVersion::v2024_03_28`: Superseded by v2025_01_17, which updates the default HTTPS client stack.


DDtKey (approved):
Thank you!

Issue #108, opened 2024-03-28
Support Azurite


tdikland:
### Feature request description

I would love to see a module to setup Azurite for testing integration with ADLS.

### Image reference

The Azurite documentation can be found here: https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio%2Cblob-storage

Dockerhub reference: https://hub.docker.com/_/microsoft-azure-storage-azurite

mervyn-mccreight:
@tdikland is this issue resolved with https://github.com/testcontainers/testcontainers-rs-modules-community/pull/334?

PR #334, opened 2025-05-04
feat: add support for Azurite docker

fulgas:
Solves  #108


src/azurite/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+            tag: "latest".to_string(),
+            ..Self::default()
+        }
+    }
+
+    /// Overrides the image tag.
+    /// Check https://mcr.microsoft.com/v2/azure-storage/azurite/tags/list to see available tags.
+    pub fn with_tag(self, tag: String) -> Self {
+        Self { tag, ..self }
+    }
```
DDtKey, line 75: We usually don't provide such method here as `testcontainers-rs` allows this customization on core level via `ImageExt` trait

DDtKey (approved):
Looks good to me!

Just one really minor comment

src/azurite/mod.rs
fulgas: I will remove it then. thk for the input

PR #348, opened 2025-05-21
feat: add arrow_flightsql module

brianheineman:


Cargo.toml
```diff
@@ -78,22 +79,18 @@ testcontainers = { version = "0.24.0" }
 
 
 [dev-dependencies]
-# Work around feature incompatibility in Chrono and arrow-arith
-# https://github.com/apache/arrow-rs/issues/7196
-# TODO: Remove when we can update databend-driver
-chrono = "= 0.4.39"
```
brianheineman, line 84: The latest version of `arrow-flight` requires a newer version of `chrono`.  I removed this workaround and updated the `databend-driver` which appears to resolve the dependency conflicts.

DDtKey (approved):
Thank you for the contribution! 👍

PR #363, opened 2025-06-27
chore: bump pinned localstack tag to 4.5

kairsas:
Default `localstack` tag 3.0 was quite outdated, 
causing S3 `put_object` operations to fail with recent versions of the AWS CLI.

Updating to the latest stable tag, 4.5, resolves the issue.


kairsas:
@DDtKey Maybe you could take a look, is this acceptable?

DDtKey:
Hi @kairsas 

We usually don’t bump image version if it doesn’t require any changes except the version/tag itself

See https://github.com/testcontainers/testcontainers-rs-modules-community/pull/138#issuecomment-2121516006 for example

I don’t mind updating it with the next major release though, so let’s keep PR active if you don’t mind

DDtKey:
> I see your point, but it doesn’t feel natural to me.

IMO, the current approach appears to be quite natural.
The primary goal here is to ensure that minor updates don't silently break the user’s code. Unfortunately, upgrading tags can easily lead to this scenario.

While I agree users should stick to specific versions on their end, it is crucial to avoid breaking things in minor patches. 
Such an action could potentially violate the semver and for users it's almost like using `latest` tag (it might be updated or no, who knows)

I'm generally open to updating images, but it should be done with appropriate versioning.

DDtKey:
Btw, could you address clippy warnings?

kairsas:
@DDtKey do you have a clue what are those failing tests? 
Looks like it should not be related to my changes..
Should I try fixing them, or will they be handled separately?

DDtKey:
Sorry for the late response, we have several flaky tests unfortunately and it's not related you your changes

It's gonna be addressed separately

PR #455, opened 2026-03-07
feat(mongo)!: switch to Mongo 8

### LINKED ISSUE #443: Update to Mongo 8

Hello, for my tests I need to test buld writes which are supported only by mongo 8+. Would it be possible to update the image?

### Image reference

_No response_

barsdeveloper:
Fixes #443

### DIFF
--- a/src/mongo/mod.rs
+++ b/src/mongo/mod.rs
@@ -4,7 +4,7 @@ use testcontainers::{
 };
 
 const NAME: &str = "mongo";
-const TAG: &str = "5.0.6";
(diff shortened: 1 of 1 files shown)


DDtKey (approved):
Thanks for the contribution!

DDtKey (changes requested):
It simply bumps the version without resolving the discussion in https://github.com/testcontainers/testcontainers-rs-modules-community/issues/443

We need to re-think how we handle that and whether we want to support older versions at all

PR #459, opened 2026-03-07
Update MariaDB

barsdeveloper:


DDtKey:
I'm closing the PR as such update doesn't require any changes in the image
You always can utilize [`with_tag`](https://docs.rs/testcontainers/0.27.0/testcontainers/core/trait.ImageExt.html#tymethod.with_tag) from `ImageExt` (we re-export `testcontainers` - https://docs.rs/testcontainers-modules/latest/testcontainers_modules/#reexports)

We typically refrain from performing tag updates unless absolutely necessary. This is because such updates could potentially be considered breaking changes for some users, and we already offer a mechanism to override the version.

PR #463, opened 2026-03-13
feat!: update rabbitmq default tag to 4.2-management

### LINKED ISSUE #462: RabbitMQ module defaults to EOL 3.8.22-management image
The RabbitMQ module currently hardcodes `3.8.22-management` as the default tag:

https://github.com/testcontainers/testcontainers-rs-modules-community/blob/main/src/rabbitmq/mod.rs#L4

```rust
const TAG: &str = "3.8.22-management";
```

RabbitMQ 3.8 has been end-of-life since July 2022. The current stable release series with community support is **4.2** (latest patch: 4.2.4, released Feb 17, 2026). See https://www.rabbitmq.com/release-information for the full support matrix.

The ready condition string may also need updating — the current check waits for:

```
Server startup complete; 4 plugins started.
```

This message may differ across major versions (e.g., different plugin count in 4.x).

### Suggestion

Bump the default tag to `4.2-management` (or `management` to track latest) and verify the ready condition still matches.

Users can work around this today with `.with_tag("4.2-management")`, but it would be nice for the default to be a supported version.

xvchris:
Updates the default RabbitMQ image tag from EOL 3.8.22-management to the current stable 4.2-management release.

Also updates the ready condition to be version-agnostic, matching "Server startup complete" instead of the specific plugin count which may vary between versions.

Fixes #462

### DIFF
--- a/src/rabbitmq/mod.rs
+++ b/src/rabbitmq/mod.rs
@@ -1,7 +1,7 @@
 use testcontainers::{core::WaitFor, Image};
 
 const NAME: &str = "rabbitmq";
-const TAG: &str = "3.8.22-management";
+const TAG: &str = "4.2-management";
 
 /// Module to work with [`RabbitMQ`] inside of tests.
 ///
@@ -48,7 +48,7 @@ impl Image for RabbitMq {
 
     fn ready_conditions(&self) -> Vec<WaitFor> {
         vec![WaitFor::message_on_stdout(
-            "Server startup complete; 4 plugins started.",
+            "Server startup complete",
         )]
     }
 }


.pr-benchmark.md
```diff
@@ -0,0 +1,39 @@
...
+  - `feat: add RustFS module`
+  - `fix: ready condition in ClickHouse`
+  - `feat!: update testcontainers to 0.27`
+
+## Diff Characteristics
+- **Files**: 通常 1-4 个文件
+- **Lines**: 小规模变更 (1-20 行，除了新模块)
+- **Tests**: 新模块会添加测试，修复通常不需要
+- **Docs**: 新模块需要 rustdocs
+- **Changelog**: 自动处理，无需手动更新
```
coconutbird: Why introduce this file?

coconutbird (changes requested):

xvchris:
Sorry about that, removed.

xvchris:
Sorry about that — removed the .pr-benchmark.md file from the commit. Only src/rabbitmq/mod.rs is changed now.

PR #472, opened 2026-05-01
feat: add Hickory DNS server module

gibbz00:


src/hickory_dns/mod.rs
```diff
@@ -0,0 +1,200 @@
+use std::borrow::Cow;
+
+use testcontainers::{
+    core::{ContainerPort, WaitFor},
+    CopyDataSource, CopyToContainer, Image,
+};
+
+const NAME: &str = "hickorydns/hickory-dns";
+const TAG: &str = "latest";
```
mervyn-mccreight: We avoid using latest as the default, because like this what the module provides can break in the future when there are updates in the image that would need changes here (e.g. the waiting condition).

What we usually do is reference a fixed release version here and only update it if we need to (e.g. in the example above where we need to change something to be compatible).

mervyn-mccreight (commented):
Thanks for the contribution! Just a small thought from my side regarding the default tag.

src/hickory_dns/mod.rs
gibbz00: Reasonable take :) Fixed it now.

DDtKey (approved):

DDtKey (approved):

DDtKey:

